PR #7, opened 2021-03-24
Add instrumentation for httpd - Apache HTTP Server

TomRoSystems:


lalitb (approved):

lalitb:
@TomRoSystems  - Should I merge the PR, asking as one of the check has failed.

TomRoSystems:
> @TomRoSystems - One of the check has failed. Should I merge the PR in case #8 fixes it ?

I'd give time until end of this week for reviews / comments as this one is bigger. And then merge on Monday. Thanks.

instrumentation/httpd/src/otel/mod_otel.cpp
```diff
@@ -0,0 +1,387 @@
...
+{
+  if (!strcasecmp(arg, "file"))
+    config.type = OtelExporterType::OSTREAM;
+  else if (!strcasecmp(arg, "otlp"))
+    config.type = OtelExporterType::OTLP;
+  else
+    return "Unknown exporter type - can be file or otlp";
+
+  return NULL;
+}
```
pyohannes, line 285: Please correct me if I'm wrong here, but I  think any kind of instrumentation shouldn't define an exporter. Exporters should only be defined by the end-user application developer. Otherwise you could only use _one_ instrumentation at a time (e. g. Apache), but never two or more (e. g. Apache and mysql) as they may conflict (e. g. intitialize two OTEL exporters).

As  rule of thumb, instrumentation should only use the OpenTelemetry API, but not the SDK.

pyohannes: Also, see for example who Go provides instrumentation for its standard HTTP server: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/example/server/server.go

The exporter is configurable via standard SDK methods, and the instrumentation is agnostic of it.

pyohannes: Just realized that in case of Apache (and nginx) we're not actually instrumenting a _library_, but and _end-user application_.

So it might be ok to specify an exporter in this case. However, it's still a bit troubling for me that user can't specify their custom exporters or configure exporters/processors without recompiling.

maxgolov (approved):

instrumentation/httpd/Makefile
```diff
@@ -0,0 +1,38 @@
+
+
```
ThomsonTan, line 2: nit: remove heading empty lines?

instrumentation/httpd/src/otel/mod_otel.cpp
```diff
@@ -0,0 +1,387 @@
...
+#include "httpd.h"
+// httpd.h has to be before those two and after opentelemetry because it #define's OK 0
+#include "http_config.h"
+#include "http_protocol.h"
+#include "mod_proxy.h"
+
+namespace
+{
+
+const std::string kSpanNamePrefix = "HTTP ";
```
ThomsonTan, line 33: Is the trailing space intended?

ThomsonTan (approved):

PR #10, opened 2021-03-29
nginx: add test matrix, release builds

seemk:
* Use only Docker for build and tests. This removed the redundant build step and allows for easier caching.
* Use only release builds for testing.
* Add more distros (Ubuntu 18.04, 20.10) along with mainline nginx (1.19.x) to the testing matrix.
* Add Docker layer caching via buildx + Github Actions cache. Shaves off 15 minutes of build time from grpc on every distro.
* Provide prebuilt libraries / artifacts for each distro and nginx configuration.
* Add CI push trigger paths to `nginx.yml` so nginx builds won't activate on unrelated PRs.
* Fix `trace.json` output file permissions for tests.


seemk:
There, good to go now. Tests found an issue with the default nginx on Ubuntu 18.04: the package manager's nginx wasn't compiled with compatibility flags.

PR #14, opened 2021-03-31
Httpd add possibility to configure custom attributes and resources

TomRoSystems:


instrumentation/httpd/opentelemetry.conf
```diff
@@ -32,4 +32,7 @@ OpenTelemetryPath /tmp/output-spans
 
 # OpenTelemetryIgnoreInbound off
 
+# OpenTelemetrySetResource service.name apache-web-server
+# OpenTelemetrySetAttribute foo bar
```
seemk, line 36: Are these attributes global, for all routes? Is it possible to set attributes for specific route?

With nginx it's possible to set these per server, route basis.

instrumentation/httpd/src/otel/opentelemetry.cpp
```diff
@@ -85,8 +85,17 @@ void initTracer()
     processor = std::make_shared<sdktrace::SimpleSpanProcessor>(std::move(exporter));
   }
 
+  opentelemetry::sdk::resource::ResourceAttributes resAttrs({{"service.name", "httpd"}});
```
seemk, line 88: From the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service):
```
[1]: MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fallback to u nknown_service: concatenated with process.executable.name, e.g. unknown_service:bash. If process.executable.name is not available, the value MUST be set to unknown_service.
```

instrumentation/httpd/src/otel/opentelemetry.cpp
```diff
@@ -85,8 +85,17 @@ void initTracer()
...
+  // add custom-configured resources
+  for(auto &it:config.resources)
```
seemk, line 90: Spacing?

instrumentation/httpd/src/otel/opentelemetry.h
```diff
@@ -54,6 +55,8 @@ struct OtelConfig
   // context propagation
   bool ignore_inbound;
   OtelPropagation propagation;
+  std::unordered_map<std::string, std::string> attribiutes;
```
seemk, line 58: Typo here

lalitb (approved):

instrumentation/httpd/opentelemetry.conf
TomRoSystems: At the moment those are global settings.

instrumentation/httpd/src/otel/opentelemetry.cpp
TomRoSystems: Yes. Thanks. Removed.

instrumentation/httpd/src/otel/opentelemetry.h
TomRoSystems: Fixed typo.

PR #25, opened 2021-05-18
feat(nginx): read OTLP endpoint from env

johanneswuerbach:
Support the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable to configure the endpoint.

This is required when the collector is running as a daemonset on each node as the collector endpoint needs to be injected by k8s.

The variable is part of the standard configuration options https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options


seemk (approved):
:+1: LGTM!

PR #26, opened 2021-05-18
feat(nginx): configurable sampler

johanneswuerbach:
Allow to configure the sampler to use parent based or ratio based sampling.

It would be great to unit test this somehow, but I haven't really done much cpp recently so would appreciate some help.


johanneswuerbach:
@seemk would you have time to take a look?

seemk (approved):
LGTM! Like that it is possible to leave sampler out from the conf file :slightly_smiling_face:

PR #32, opened 2021-06-01
httpd add CMake build and CI

TomRoSystems:


instrumentation/httpd/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+  ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+  "/usr/include/apache2"
+  "/usr/include/apr-1.0"
+)
+
+target_link_libraries(otel_httpd_module
+  PRIVATE
+  ${OPENTELEMETRY_CPP_LIBRARIES}
+  gRPC::grpc++
+#  Thrift::thrift
```
seemk, line 38: Thrift is searched for, but is not used?

instrumentation/httpd/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+
+set_target_properties(otel_httpd_module PROPERTIES
+  PREFIX ""
+)
+
+target_include_directories(otel_httpd_module
+  PRIVATE
+  ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+  "/usr/include/apache2"
+  "/usr/include/apr-1.0"
```
seemk, line 31: I think Apache headers should be searched for to omit hardcoding of this search path.

E.g. it should be possible to do `-DCMAKE_PREFIX_PATH="/usr/local/customapachedir"` and CMake will pick up the include paths from it.

A simple example how to to write such a CMake script can be found at https://github.com/seemk/FastHumanDetection/blob/master/modules/FindKinectV2.cmake (unrelated project, just a simple example how a header file can be searched for)

instrumentation/httpd/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+find_package(CURL REQUIRED)
+find_package(Thrift REQUIRED)
+
+add_library(otel_httpd_module SHARED
+  src/otel/mod_otel.cpp
+  src/otel/opentelemetry.cpp
+)
+
+target_compile_options(otel_httpd_module
+  PRIVATE -Wall -Wextra -fpermissive
```
seemk, line 18: Imho we should not allow for `-fpermissive` and fix the underlying issues instead

instrumentation/httpd/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.12)
+
+project(opentelemetry-httpd)
+
+find_package(opentelemetry-cpp REQUIRED)
+find_package(Threads REQUIRED)
+find_package(protobuf REQUIRED)
+find_package(gRPC REQUIRED)
+find_package(CURL REQUIRED)
```
seemk, line 9: Is curl needed?

instrumentation/httpd/src/otel/opentelemetry.cpp
```diff
@@ -30,17 +30,17 @@ namespace httpd_otel
...
+const opentelemetry::v0::nostd::string_view kAttrHTTPMethod             = "http.method";
+const opentelemetry::v0::nostd::string_view kAttrHTTPScheme             = "http.scheme";
+const opentelemetry::v0::nostd::string_view kAttrHTTPHost               = "http.host";
+const opentelemetry::v0::nostd::string_view kAttrHTTPTarget             = "http.target";
+const opentelemetry::v0::nostd::string_view kAttrHTTPUrl                = "http.url";
+const opentelemetry::v0::nostd::string_view kAttrHTTPFlavor             = "http.flavor";
+const opentelemetry::v0::nostd::string_view kAttrHTTPClientIP           = "http.client_ip";
+const opentelemetry::v0::nostd::string_view kAttrNETPeerIP              = "net.peer.ip";
+const opentelemetry::v0::nostd::string_view kAttrHTTPStatusCode         = "http.status_code";
+const opentelemetry::v0::nostd::string_view kAttrHTTPResponseContentLen = "http.response_content_length";
```
seemk, line 43: My hunch is that the explicit namespaces are not needed here and the issue is somewhere else, can you double check?

instrumentation/httpd/CMakeLists.txt
TomRoSystems: Yup. Fixed with `reinterpret_cast`

instrumentation/httpd/src/otel/opentelemetry.cpp
TomRoSystems: You're right. I've removed those.

instrumentation/httpd/CMakeLists.txt
TomRoSystems: Yes without it I've got an error: `... links to target "CURL::libcurl" but the target was not found.`

instrumentation/httpd/CMakeLists.txt
TomRoSystems: Yes, as at the moment I'm not using Jaeger exporter yet. And it compiles - you can check results as I've added CMake to continuous integration to GitHub actions.

instrumentation/httpd/src/otel/opentelemetry.h
```diff
@@ -85,8 +86,8 @@ struct HttpdEndSpanAttributes
 // to be called explictly inside Destruct method
 struct ExtraRequestData
 {  // context which we pass
-  nostd::unique_ptr<opentelemetry::context::Token> token;
-  nostd::shared_ptr<opentelemetry::v0::trace::Span> span;
+  opentelemetry::v0::nostd::unique_ptr<opentelemetry::context::Token> token;
```
seemk, line 89: Is the `v0` prefix needed?

TomRoSystems: Prefix removed.

seemk (approved):

PR #45, opened 2021-07-13
feat(nginx): opentelemetry-cpp v1.0.0-rc3

johanneswuerbach:
Bump to https://github.com/open-telemetry/opentelemetry-cpp/releases/tag/v1.0.0-rc3

### DIFF
--- a/instrumentation/nginx/CMakeLists.txt
+++ b/instrumentation/nginx/CMakeLists.txt
@@ -7,6 +7,7 @@ find_package(Threads REQUIRED)
 find_package(protobuf REQUIRED)
 find_package(gRPC REQUIRED)
 find_package(CURL REQUIRED)
+find_package(nlohmann_json REQUIRED)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/nginx.cmake)
 
@@ -24,7 +25,7 @@ add_library(otel_ngx_module SHARED
 target_compile_options(otel_ngx_module
   PRIVATE -Wall -Wextra
 )
-  
+
 install(TARGETS otel_ngx_module DESTINATION ".")
 
 # Omit the lib prefix to be more in line with nginx module naming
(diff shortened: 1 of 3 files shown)


instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -17,7 +17,7 @@ extern ngx_module_t otel_ngx_module;
 #include "nginx_utils.h"
 #include "propagate.h"
 #include <opentelemetry/context/context.h>
-#include <opentelemetry/exporters/otlp/otlp_exporter.h>
+#include <opentelemetry/exporters/otlp/otlp_grpc_exporter.h>
```
johanneswuerbach, line 20: This was due to https://github.com/open-telemetry/opentelemetry-cpp/pull/810

seemk (approved):
Thanks! That was fast :smile:

instrumentation/nginx/CMakeLists.txt
```diff
@@ -44,4 +45,5 @@ target_link_libraries(otel_ngx_module
   ${OPENTELEMETRY_CPP_LIBRARIES}
   gRPC::grpc++
   CURL::libcurl
+  nlohmann_json::nlohmann_json
```
johanneswuerbach, line 48: This is coming from https://github.com/open-telemetry/opentelemetry-cpp/pull/810, not sure if we should attempt to drop this as `OTLP/HTTP+JSON exporter` isn't used at the moment.

seemk: Does it build when not using `find_package(nlohmann_json)` and not linking to it?

johanneswuerbach: Nope :-(

seemk: Otel CPP external dependencies are starting to get a bit out of hand, wish it had none and would be distributed as a single library file. :smile: 

Anyway, thanks for the PR!

PR #49, opened 2021-08-23
nginx: rename and document context accessor variable

seemk:
The trace context can now be accessed via `opentelemetry_context_traceparent` variable.

Updated docs and added a test for it.

### DIFF
--- a/instrumentation/nginx/README.md
+++ b/instrumentation/nginx/README.md
@@ -170,7 +170,7 @@ The following nginx variables are set by the instrumentation:
 - `opentelemetry_context_traceparent` - [W3C trace
   context](https://www.w3.org/TR/trace-context/#trace-context-http-headers-format), e.g.: `00-0af7651916cd43dd8448eb211c80319c-b9c7c989f97918e1-01`
 - `opentelemetry_context_b3` - Trace context in the [B3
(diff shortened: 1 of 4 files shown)


instrumentation/nginx/README.md
```diff
@@ -163,6 +163,21 @@ List of exported attributes and their corresponding nginx variables if applicabl
...
 - `http.server_name` - From the `server_name` directive
 
+## nginx variables
+
+The following nginx variables are set by the instrumentation:
+
+- `opentelemetry_context_traceparent` - [W3C trace
+  context](https://www.w3.org/TR/trace-context/#trace-context-http-headers-format), e.g.: `00-0af7651916cd43dd8448eb211c80319c-b9c7c989f97918e1-01`
+- `opentelemetry_context_b3` - Trace context in the [B3
+  format](https://github.com/openzipkin/b3-propagation#single-header). Only set when using `opentelemetry propagate b3`.
```
ThomsonTan, line 173:
```suggestion
  format](https://github.com/openzipkin/b3-propagation#single-header). Only set when using `opentelemetry propagator b3`.
```

seemk: Thanks! Fixed the typo in README: `opentelemetry_propagate`, although `opentelemetry_propagator` would be a better option indeed :thinking: But don't want to rename or add an alias with this PR

ThomsonTan (approved):

PR #50, opened 2021-08-31
Add support for alpine to nginx

dmathieu:
This adds support for alpine to the nginx plugin (there was an issue with the `Protobuf` package name, which needs to be capitalized), testing that distribution in CI.

### DIFF
--- a/.github/workflows/nginx.yml
+++ b/.github/workflows/nginx.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


dmathieu:
So, the thing is, nginx in APK doesn't have `--with-compat` 😢 

See https://gitlab.alpinelinux.org/alpine/aports/-/issues/12375
And https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/18945

Though, based on the comments in that merge request, it doesn't look like the APK maintainers are too keen on adding that option.

I'll open another PR for the `s/protobuf/Protobuf`. I'm happy to hear proposals for this one.

PR #52, opened 2021-09-10
Setup nginx config file for dynamic module building

dmathieu:
This sets up a `config` file so nginx knows how to use this module as a dynamic one.
See https://www.nginx.com/resources/wiki/extending/new_config/

With this change, passing the option

```
--add-dynamic-module=/path/to/opentelemetry-cpp-contrib/instrumentation/nginx"
```

to when configuring nginx allows building this plugin as well.

### DIFF
--- a/instrumentation/nginx/config
+++ b/instrumentation/nginx/config
@@ -1,0 +1,24 @@
+ngx_addon_name=opentelemetry_nginx
+ngx_module_type=HTTP
+ngx_module_name=$ngx_addon_name
+ngx_module_incs=
+ngx_module_deps=" \
+  $ngx_addon_dir/src/agent_config.h \
+  $ngx_addon_dir/src/location_config.h \
+  $ngx_addon_dir/src/nginx_config.h \
+  $ngx_addon_dir/src/nginx_utils.h \
+  $ngx_addon_dir/src/propagate.h \
+  $ngx_addon_dir/src/script.h \
+  $ngx_addon_dir/src/toml.h \
+  $ngx_addon_dir/src/trace_context.h \
+"
+ngx_module_srcs=" \
+  $ngx_addon_dir/src/agent_config.cpp \
+  $ngx_addon_dir/src/nginx_config.cpp \
+  $ngx_addon_dir/src/otel_ngx_module.cpp \
+  $ngx_addon_dir/src/propagate.cpp \
+  $ngx_addon_dir/src/script.cpp \
+  $ngx_addon_dir/src/trace_context.cpp \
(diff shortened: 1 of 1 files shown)


dmathieu:
Hi @TomRoSystems @seemk,
Is there anything I can do to help you review this PR?

seemk:
> Hi @TomRoSystems @seemk,
> Is there anything I can do to help you review this PR?

Hey, thanks for the ping :) I was away for 2 weeks and just came back today.

With this config file, how are for example OpenTelemetry CPP include paths handled? Are they passed to the nginx config command line?

dmathieu:
Yes, the include paths are passed to the nginx config command line.
As an example, you can see my draft PR adding this module to ingress-nginx: https://github.com/kubernetes/ingress-nginx/pull/7621/files#diff-d96b4dbdd33ef5b9f1b1c7ce45ef0a515f55a55f130318c1af397e731a42d232R674

seemk:
If I understand correctly currently it's looking for the include paths in the system path? So if Otel CPP is installed elsewhere, `ngx_module_incs` would need to be set in the config file? And the same goes for linking?

Not that it matters much, could very well live with this for now, just trying to understand it :smile:

dmathieu:
I'm not really an expert at C compilation. I've adapted this config file from the one in opentracing: https://github.com/opentracing-contrib/nginx-opentracing/blob/master/opentracing/config
I'm happy to add any config that you will think would be better.

seemk:
In your `ingress-nginx` PR does it work like this? My hunch is that all the OpenTelemetry libs and libstdc++ would need to be linked to the module at the config file as well: `ngx_module_libs="-lstdc++ -libopentelemetry_trace ..."`. 

However OpenTelemetry CPP consists of a huge list of libraries that all need to be linked, all the `libabsl_*`, `libopentelemetry_*`.
This was something I had issues with when initially trying to build it via nginx's build system and opted for a CMake build with a workaround (e.g. https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/nginx/src/otel_ngx_module_modules.c) that mimics what the nginx build system does.

dmathieu:
I haven't been able to do extensive testing of the feature, and sending data yet. But yes, with the library being built in my ingress-nginx, I am able to build the library, and start nginx with the module configured.

seemk:
Ok, weird. It might be because of system paths then, let's merge it for now if it helps with `ingress-nginx`, would love to see it used there. Thanks!

seemk (approved):

PR #58, opened 2021-09-23
nginx: link the module against libstd++ and the (used) opentelemetry libraries in the nginx build

tobiasstadler:


seemk:
Thanks, are these libraries enough? See the discussion at https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/52

tobiasstadler:
Yes, I think so. At least it is working for me that way (building opentelemtry as shared libraries)

tobiasstadler:
If (for other use cases) there is something missing we can still add it later

seemk (approved):

tobiasstadler:
Thank You!

PR #59, opened 2021-09-25
nginx: Expose trace id and span id as nginx vars

tobiasstadler:


instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -145,6 +167,74 @@ OtelGetTraceContextVar(ngx_http_request_t* req, ngx_http_variable_value_t* v, ui
...
+      NGX_LOG_ERR, req->connection->log, 0,
+      "Unable to get trace context when getting trace id");
+    return NGX_OK;
+  }
+
+  trace::SpanContext spanContext = traceContext->request_span->GetContext();
+
+  if (spanContext.IsValid()) {
+    constexpr int len = 2 * trace::TraceId::kSize;
+    char* data = (char*)ngx_palloc(traceContext->request->pool, len);
```
seemk, line 185: You could also use `req->pool` directly here.
Also `ngx_palloc` might return `NULL`

seemk (commented):
Thanks! Minor notes regarding `ngx_palloc`. Are you ok with adding a test case for trace, span ID as well?

tobiasstadler:
> Thanks! Minor notes regarding `ngx_palloc`. Are you ok with adding a test case for trace, span ID as well?

I am unfamiliar with the testing framework, but I will give it a try.

seemk (approved):
Thanks a lot!

tobiasstadler:
Thank You!

PR #60, opened 2021-09-27
nginx: Added option to distrust the incoming spans

tobiasstadler:


seemk (approved):
Thanks, like the idea :+1:

tobiasstadler:
Thank You!

PR #61, opened 2021-09-27
nginx: Enabled operation_name to be specifed inside a server block

tobiasstadler:


instrumentation/nginx/src/script.h
```diff
@@ -26,6 +26,10 @@ struct NgxCompiledScript {
 
     return true;
   }
+
+  bool isValid() {
```
seemk, line 30: Perhaps nitpicking, but maybe `isEmpty`? `isValid` has more of a "is in an error state" vibe :smile:

tobiasstadler: Yes, `IsEmpty` sounds better

seemk (approved):
Thanks :+1:

tobiasstadler:
Thank You!

Issue #16, opened 2021-04-14
Docker image on docker hub


teuno:
Is it possible to have an docker image on dockerhub (or somewhere else)? 
I am building an example of a system monitored with opentelemetry, but struggeling how to build this project, because I do not have that much experience with building Nginx from source (never done that). 

Thanks in advance, both it it happens and if it does not happen.

seemk:
Thanks! I'll look into publishing to Docker hub, I agree building it is not a trivial matter due to requiring building of both gRPC and opentelemetry-cpp.

What platforms did you try to build it on?

We don't have the releases process set up yet (#11), but the CI does build prebuilt binaries on Ubuntu 18.04, 20.04 and 20.10 for both nginx 1.18 and 1.19 which you can grab here: https://github.com/open-telemetry/opentelemetry-cpp-contrib/actions/runs/705078540

I can add additional targeted platforms as well.

teuno:
Hi, atm I have followed https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#modules . Got a working nginx without the grpc and otlp, but kinda stuck after that (the dependencies for building part in https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/nginx ). I build on ubuntu 18.04, so that's one that you already have.

Do you know which --add(-dynamic)-module to add?

seemk:
Just to get an overview of your setup:
You are building nginx yourself? If yes, then I think you only need to add `--with-compat` flag. See https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/nginx#troubleshooting

Are you using the prebuilt `.so`?

If you are building nginx you don't need to add any `-add-*-module` for instrumentation module to work, in fact it should work with nginx versions out of the box (with the exception of nginx 1.14 on ubuntu 18.04 default apt - you can get a newer one from nginx packages: https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#installing-a-prebuilt-debian-package-from-an-os-repository)

Then if you have nginx and the instrumentation module `.so` you can add it to nginx config. Example - https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/nginx#usage

You can check the nginx flags with `nginx -V`, it should have `--with-compat`.

teuno:
[...]

PR #62, opened 2021-10-06
Add download location to usage docs

phillipuniverse:
I wanted to have a small conversation about the best place to include this. I came across this [in this issue comment](https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/16#issuecomment-820194434) and thought it would be better to formalize into the README to highlight that you didn't have to build it from scratch.

It would be nice if this was a binary from the 'Releases' section to make it even easier to download.

Finally, while I get that this is just pre-built from Ubuntu, is there any way to make it compatible with Debian as well? I have a `Dockerfile` that does this:

```
FROM nginx:1.21.3  # this is based on Debian
ADD modules/* /etc/nginx/modules
```

Then reference the .so file in a `load_module` directive. I get this error:

```
nginx: [emerg] dlopen() "/etc/nginx/modules/otel_ngx_module.so" failed (/lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /etc/nginx/modules/otel_ngx_module.so)) in /etc/nginx/common-outer-config.nginx.conf:1
```

### DIFF
--- a/instrumentation/nginx/README.md
+++ b/instrumentation/nginx/README.md
@@ -17,6 +17,7 @@ Supported propagation types:
 
 Additional platforms and/or versions coming soon.
 
+
 ## Dependencies (for building)
 
 1. [gRPC](https://github.com/grpc/grpc) - currently the only supported exporter is OTLP. This requirement will be lifted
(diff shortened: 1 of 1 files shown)


seemk:
Hey, thanks for the contribution. I'm currently not very happy with the build process, mainly due to the logistics involved to build everything. But as OpenTelemetry C++ is now available via [Conan](https://conan.io/center/search/opentelemetry-cpp), I want to move all the builds there and provide additional prebuilt binaries.

seemk (approved):

PR #73, opened 2021-11-24
Populate the net.peer.ip and net.peer.port atttributes

tobiasstadler:


instrumentation/nginx/test/instrumentation/test/instrumentation_test.exs
```diff
@@ -154,6 +154,8 @@ defmodule InstrumentationTest do
     assert status == 200
 
     assert attrib(span, "net.host.port") == "8000"
+    assert attrib(span, "net.peer.ip") != nil
```
seemk, line 157:
```suggestion
    assert attrib(span, "net.peer.ip") =~ ~r/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/
```

instrumentation/nginx/test/instrumentation/test/instrumentation_test.exs
```diff
@@ -154,6 +154,8 @@ defmodule InstrumentationTest do
...
+    assert attrib(span, "net.peer.port") != nil
```
seemk, line 158:
```suggestion
    assert attrib(span, "net.peer.port") =~ ~r/\d+/
```

seemk (approved):
Thanks! Perhaps we should also check for the shape of the attribute? Added my suggestions 🙂

tobiasstadler:
Thank You!

PR #75, opened 2021-11-25
nginx: Populate http.request.header.* and http.response.header.*

tobiasstadler:


instrumentation/nginx/README.md
```diff
@@ -159,6 +159,30 @@ be started. The default propagator is W3C.
...
+
+Enables the capturing of request and response headers. (default: disabled).
+
+- **required**: `false`
+- **syntax**: `opentelemetry_capture_headers on|off`
+- **block**: `http`, `server`, `location`
+
+### `opentelemetry_sensitive_header_names`
+
+Sets the caputured header value to `[REDACTED]` for all headers where the name matches the given regex (case insensitive).
```
seemk, line 172:
```suggestion
Sets the captured header value to `[REDACTED]` for all headers where the name matches the given regex (case insensitive).
```

instrumentation/nginx/README.md
```diff
@@ -159,6 +159,30 @@ be started. The default propagator is W3C.
...
+
+- **required**: `false`
+- **syntax**: `opentelemetry_sensitive_header_names <regex>`
+- **block**: `http`, `server`, `location`
+
+### `opentelemetry_sensitive_header_values`
+
+Sets the caputured header value to `[REDACTED]` for all headers where the value matches the given regex (case insensitive).
```
seemk, line 180:
```suggestion
Sets the captured header value to `[REDACTED]` for all headers where the value matches the given regex (case insensitive).
```

instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -65,6 +66,62 @@ nostd::string_view NgxHttpFlavor(ngx_http_request_t* req) {
...
+#endif
+
+      nostd::string_view value;
+      if (sensitiveHeader) {
+        value = "[REDACTED]";
+      } else {
+        value = FromNgxString(header[i].value);
+      }
+
+      span->SetAttribute((const char*)key, nostd::span<const nostd::string_view>(&value, 1));
```
seemk, line 120: minor: the `strlen` used in the construction of `nostd::string_view` can be avoided if `key` length is passed in here (and it would get rid of the need for a zero terminator).

instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -65,6 +66,62 @@ nostd::string_view NgxHttpFlavor(ngx_http_request_t* req) {
...
+
+    dst[i] = ch;
+  }
+
+  dst[str.len] = '\0';
+}
+
+static void OtelCaptureHeaders(nostd::shared_ptr<opentelemetry::trace::Span> span, ngx_str_t keyPrefix, ngx_list_t *headers,
+                        ngx_regex_t *sensitiveHeaderNames, ngx_regex_t *sensitiveHeaderValues,
+                        std::vector<ngx_table_elt_t*> excludedHeaders = {}) {
```
seemk, line 86: Would passing `std::vector` passing by value here create an unnecessary heap alloc / free? Can't say without testing, but perhaps `std::array` would work here to avoid the potential allocation?

seemk (commented):
Thanks :slightly_smiling_face: some minor comments

instrumentation/nginx/src/otel_ngx_module.cpp
tobiasstadler: Using std::array ist not possible, because the array size is part if the type (Maybe if I convert `OtelCaptureHeaders` to a template method). I used nosed::span now.

seemk: Right, thanks :smile:

tobiasstadler:
Thank You!

PR #76, opened 2021-11-30
nginx: net.host.port and net.peer.port should be ints

tobiasstadler:
See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md

### DIFF
--- a/instrumentation/nginx/src/script.h
+++ b/instrumentation/nginx/src/script.h
@@ -32,23 +32,28 @@ struct NgxCompiledScript {
   }
 };
 
+enum ScriptAttributeType {
+  ScriptAttributeInt,
+  ScriptAttributeString,
+};
+
 struct ScriptAttributeDeclaration {
   ScriptAttributeDeclaration(
-    opentelemetry::nostd::string_view attribute, opentelemetry::nostd::string_view script, bool convertToInt = false)
-    : attribute(ToNgxString(attribute)), script(ToNgxString(script)), convertToInt(convertToInt) {}
+    opentelemetry::nostd::string_view attribute, opentelemetry::nostd::string_view script, ScriptAttributeType type = ScriptAttributeString)
+    : attribute(ToNgxString(attribute)), script(ToNgxString(script)), type(type) {}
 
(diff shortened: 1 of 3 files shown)


instrumentation/nginx/src/script.h
```diff
@@ -34,19 +34,21 @@ struct NgxCompiledScript {
 
 struct ScriptAttributeDeclaration {
   ScriptAttributeDeclaration(
-    opentelemetry::nostd::string_view attribute, opentelemetry::nostd::string_view script)
-    : attribute(ToNgxString(attribute)), script(ToNgxString(script)) {}
+    opentelemetry::nostd::string_view attribute, opentelemetry::nostd::string_view script, bool convertToInt = false)
```
seemk, line 37: Nice catch :smile: 

Would an enum make more sense here (and be a bit more readable when reading the declarations list)?

E.g. 
```
  {"http.scheme", "$scheme"},
  {"net.host.port", "$server_port", ScriptAttributeType_Int},
  {"net.peer.ip", "$remote_addr"},
  {"net.peer.port", "$remote_port", ScriptAttributeType_Int},
```

Then the constructor would be something like
```
ScriptAttributeDeclaration(
    opentelemetry::nostd::string_view attribute, opentelemetry::nostd::string_view script, ScriptAttributeType = ScriptAttributeType_String)
```

seemk (approved):
Thanks! Although I wonder why the traces in tests still have the values as strings :thinking:

tobiasstadler:
> Although I wonder why the traces in tests still have the values as strings 🤔

They don't, sort of. The actually appear as `%{"key" => "net.host.port", "value" => %{"intValue" => "8000"}}`, but the `Map.values(val)` in `attrib` removes the int information.

seemk:
> > Although I wonder why the traces in tests still have the values as strings thinking
> 
> They don't, sort of. The actually appear as `%{"key" => "net.host.port", "value" => %{"intValue" => "8000"}}`, but the `Map.values(val)` in `attrib` removes the int information.

Ah, thanks!

tobiasstadler:
Thank You!

PR #94, opened 2021-12-23
nginx: Create map to store requests and their trace contexts

### LINKED ISSUE #89: Nginx worker thread dies if log_subrequest is enabled
**Describe your environment**
I'm using the `iquirino91/nginx-opentelemetry` Docker image from @iquirino. See #81 and the Dockerfile [here](https://github.com/iquirino/nginx-opentelemetry/blob/7cf899d0835c4c86cabc2d5ced44224d29b39dfd/Dockerfile).

**Steps to reproduce**
The steps below assume you have Docker installed. If you do not, you are welcome to just use the `nginx.conf` with `log_subrequest` and try to reproduce the issue.

Create a new folder named `bug` and put the following files into it.

---------

`otel-nginx.toml`
```TOML
exporter = "otlp"
processor = "simple"

[exporters.otlp]
host = "otel"
port = 4317

[service]
name = "nginx-test-crash"
```
`nginx.conf`
```Conf
load_module /etc/nginx/modules/otel_ngx_module.so;

worker_processes   auto;
error_log          /var/log/nginx/error.log debug;
working_directory  /tmp/cores/;
worker_rlimit_core 500M;

events {
}
http {
  opentelemetry_config /conf/otel-nginx.toml;
  log_subrequest on;

  server {
    opentelemetry on;
    listen 80;
    server_name otel_example;

    location = / {
      opentelemetry on;

      auth_request /auth;
    }

    location = /auth {
    }

  }
}
```
---------

1. Run the following command in the `bug` folder with the files.
```
docker run \
    -v $(pwd)/nginx.conf:/etc/nginx/nginx.conf \
    -v $(pwd)/otel-nginx.toml:/conf/otel-nginx.toml \
    -v $(pwd):/tmp/cores \
    --rm \
    --name nginx-otel-bug \
    iquirino91/nginx-opentelemetry@sha256:008914fd4d4aa8e060e14d7a2af56e48418d95832251473a8cffb5c348681703
```
2. In another terminal, `exec` into Nginx with `docker exec -it nginx-otel-bug /bin/sh`.
```
# apk add curl
# curl localhost:80
```
3. Observe the crash in your original terminal.
```
2021/12/23 11:38:08 [error] 8#0: *1 open() "/etc/nginx/html/auth" failed (2: No such file or directory), client: 127.0.0.1, server: otel_example, request: "GET / HTTP/1.1", subrequest: "/auth", host: "localhost"
127.0.0.1 - - [23/Dec/2021:11:38:08 +0000] "GET / HTTP/1.1" 404 0 "-" "curl/7.80.0"
2021/12/23 11:38:08 [error] 8#0: *1 auth request unexpected status: 404, client: 127.0.0.1, server: otel_example, request: "GET / HTTP/1.1", host: "localhost"
[Error] File: /opentelemetry-cpp/exporters/otlp/src/otlp_grpc_exporter.cc:132[OTLP TRACE GRPC Exporter] Export() failed: DNS resolution failed for service: otel:4317
127.0.0.1 - - [23/Dec/2021:11:38:08 +0000] "GET / HTTP/1.1" 500 177 "-" "curl/7.80.0"
2021/12/23 11:38:09 [notice] 1#0: signal 17 (SIGCHLD) received from 8
2021/12/23 11:38:09 [alert] 1#0: worker process 8 exited on signal 11
2021/12/23 11:38:09 [notice] 1#0: start worker process 62
2021/12/23 11:38:09 [notice] 1#0: signal 29 (SIGIO) received
```
4. We can use `gdb` to check the crash in the Nginx terminal. The dump should be `/tmp/cores/core`. You can also just use your system since `/tmp/cores` is mounted to the working directory so the `core` file should be on your local disk as well. The `SetAttribute` call in `FinishNgxSpan` is causing the crash.
```
# apk add gdb
# gdb /etc/nginx/nginx /tmp/cores/core
(gdb) bt
#0  0x00007fac054e69b1 in opentelemetry::v1::sdk::trace::Span::SetAttribute(opentelemetry::v1::nostd::string_view, absl::lts_20210324::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&) () from /etc/nginx/modules/otel_ngx_module.so
#1  0x00007fac054b9a5b in FinishNgxSpan(ngx_http_request_s*) () from /etc/nginx/modules/otel_ngx_module.so
#2  0x000055655d74482d in ngx_http_log_request (r=r@entry=0x7fac04a49370) at src/http/ngx_http_request.c:3753
#3  0x000055655d74613f in ngx_http_free_request (r=r@entry=0x7fac04a49370, rc=rc@entry=0) at src/http/ngx_http_request.c:3699
#4  0x000055655d746228 in ngx_http_close_request (r=r@entry=0x7fac04a49370, rc=rc@entry=0) at src/http/ngx_http_request.c:3645
#5  0x000055655d746c10 in ngx_http_finalize_connection (r=r@entry=0x7fac04a49370) at src/http/ngx_http_request.c:2788
#6  0x000055655d7474a7 in ngx_http_finalize_request (r=r@entry=0x7fac04a49370, rc=<optimized out>) at src/http/ngx_http_request.c:2661
#7  0x000055655d7471ad in ngx_http_finalize_request (r=r@entry=0x7fac04a49370, rc=500) at src/http/ngx_http_request.c:2547
#8  0x000055655d741a49 in ngx_http_core_access_phase (r=0x7fac04a49370, ph=0x7fac04a29b38) at src/http/ngx_http_core_module.c:1148
#9  0x000055655d73cf77 in ngx_http_core_run_phases (r=0x7fac04a49370) at src/http/ngx_http_core_module.c:875
#10 0x000055655d745e99 in ngx_http_run_posted_requests (c=c@entry=0x7fac04a035a0) at src/http/ngx_http_request.c:2450
#11 0x000055655d748536 in ngx_http_process_request_headers (rev=rev@entry=0x7fac049f6490) at src/http/ngx_http_request.c:1521
#12 0x000055655d7488ca in ngx_http_process_request_line (rev=rev@entry=0x7fac049f6490) at src/http/ngx_http_request.c:1165
#13 0x000055655d748a8c in ngx_http_wait_request_handler (rev=0x7fac049f6490) at src/http/ngx_http_request.c:503
#14 0x000055655d72f91a in ngx_epoll_process_events (cycle=<optimized out>, timer=<optimized out>, flags=<optimized out>) at src/event/modules/ngx_epoll_module.c:901
#15 0x000055655d725e6b in ngx_process_events_and_timers (cycle=cycle@entry=0x7fac0554b200) at src/event/ngx_event.c:247
#16 0x000055655d72dc2d in ngx_worker_process_cycle (cycle=0x7fac0554b200, data=<optimized out>) at src/os/unix/ngx_process_cycle.c:721
#17 0x000055655d72c290 in ngx_spawn_process (cycle=cycle@entry=0x7fac0554b200, proc=proc@entry=0x55655d72db52 <ngx_worker_process_cycle>, data=data@entry=0x0, name=name@entry=0x55655d7ab53d "worker process", respawn=respawn@entry=-3)
    at src/os/unix/ngx_process.c:199
#18 0x000055655d72ce7e in ngx_start_worker_processes (cycle=cycle@entry=0x7fac0554b200, n=8, type=type@entry=-3) at src/os/unix/ngx_process_cycle.c:344
#19 0x000055655d72e336 in ngx_master_process_cycle (cycle=cycle@entry=0x7fac0554b200) at src/os/unix/ngx_process_cycle.c:130
#20 0x000055655d70623d in main (argc=<optimized out>, argv=<optimized out>) at src/core/nginx.c:383
```

**What is the expected behavior?**
Using `log_subrequest on;` should not cause the worker thread to crash.

**What is the actual behavior?**
The worker thread crashes with `log_subrequest on;` in the Nginx configuration.

rcjsuen:
Nginx requests share variables with their subrequests so we should not create contexts without first checking if it is a subrequest. As the variable is shared between the requests, we can instead create a map that will record which request matches to which context so that we can retrieve the right tracing context when `FinishNgxSpan` is called.

This fixes #89.

### DIFF
--- a/instrumentation/nginx/src/otel_ngx_module.cpp
+++ b/instrumentation/nginx/src/otel_ngx_module.cpp
@@ -331,7 +331,7 @@ void TraceContextCleanup(void* data) {
 
 void RequestContextMapCleanup(void* data) {
   std::unordered_map<ngx_http_request_t*, TraceContext*>* map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)data;
-  map->clear();
+  map->~unordered_map();
 }
 
 nostd::string_view GetOperationName(ngx_http_request_t* req) {
--- a/instrumentation/nginx/src/otel_ngx_module.cpp
+++ b/instrumentation/nginx/src/otel_ngx_module.cpp
@@ -178,7 +178,8 @@ TraceContext* GetTraceContext(ngx_http_request_t* req) {
     return nullptr;
   }
 
-  return (TraceContext*)val->data;
+  std::unordered_map<ngx_http_request_t*, TraceContext*>* map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)val->data;
+  return map->at(req);
 }
 
 nostd::string_view WithoutOtelVarPrefix(ngx_str_t value) {
@@ -328,6 +329,11 @@ void TraceContextCleanup(void* data) {
   context->~TraceContext();
 }
 
+void RequestContextMapCleanup(void* data) {
+  std::unordered_map<ngx_http_request_t*, TraceContext*>* map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)data;
+  map->clear();
+}
+
 nostd::string_view GetOperationName(ngx_http_request_t* req) {
   OtelNgxLocationConf* locationConf = GetOtelLocationConf(req);
 
@@ -365,6 +371,28 @@ static bool IsOtelEnabled(ngx_http_request_t* req) {
   return locConf->enabled;
 }
 
+TraceContext* CreateTraceContext(ngx_http_request_t* req, ngx_http_variable_value_t* val) {
+  ngx_pool_cleanup_t* cleanup = ngx_pool_cleanup_add(req->pool, sizeof(TraceContext));
+  TraceContext* context = (TraceContext*)cleanup->data;
+  new (context) TraceContext(req);
+  cleanup->handler = TraceContextCleanup;
+
+  std::unordered_map<ngx_http_request_t*, TraceContext*>* map;
+  if (req->parent) {
+    // Subrequests will already have the map created so just retrieve it
+    map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)val->data;
+  } else {
+    ngx_pool_cleanup_t* cleanup = ngx_pool_cleanup_add(req->pool, sizeof(std::unordered_map<ngx_http_request_t*, TraceContext*>));
+    map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)cleanup->data;
+    new (map) std::unordered_map<ngx_http_request_t*, TraceContext*>();
+    cleanup->handler = RequestContextMapCleanup;
+    val->data = (unsigned char*)cleanup->data;
+    val->len = sizeof(std::unordered_map<ngx_http_request_t*, TraceContext*>);
+  }
+  map->insert({req, context});
+  return context;
+}
+
 ngx_int_t StartNgxSpan(ngx_http_request_t* req) {
   if (!IsOtelEnabled(req)) {
     return NGX_DECLINED;
@@ -377,13 +405,7 @@ ngx_int_t StartNgxSpan(ngx_http_request_t* req) {
     return NGX_DECLINED;
   }
 
-  ngx_pool_cleanup_t* cleanup = ngx_pool_cleanup_add(req->pool, sizeof(TraceContext));
-  TraceContext* context = (TraceContext*)cleanup->data;
-  new (context) TraceContext(req);
-  cleanup->handler = TraceContextCleanup;
-
-  val->data = (unsigned char*)cleanup->data;
-  val->len = sizeof(TraceContext);
+  TraceContext* context = CreateTraceContext(req, val);
 
   OtelCarrier carrier{req, context};
   opentelemetry::context::Context incomingContext;
--- a/instrumentation/nginx/src/otel_ngx_module.cpp
+++ b/instrumentation/nginx/src/otel_ngx_module.cpp
@@ -179,15 +179,12 @@ TraceContext* GetTraceContext(ngx_http_request_t* req) {
   }
 
   std::unordered_map<ngx_http_request_t*, TraceContext*>* map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)val->data;
-  unsigned count = map->count(req);
-  if (count == 1) {
-    return map->at(req);
-  } else if (count > 1) {
-    ngx_log_error(NGX_LOG_ERR, req->connection->log, 0, "More than one TraceContext found for a request");
-    return nullptr;
-  }
-  ngx_log_error(NGX_LOG_ERR, req->connection->log, 0, "TraceContext not found");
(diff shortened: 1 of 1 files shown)


instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -328,6 +329,11 @@ void TraceContextCleanup(void* data) {
   context->~TraceContext();
 }
 
+void RequestContextMapCleanup(void* data) {
```
seemk, line 332: Should this instead call the destructor? I'm not sure whether `clear` deallocates anything the map has allocated internally.

seemk (commented):
Hi, thanks for the PR and sorry for the delay!

instrumentation/nginx/src/otel_ngx_module.cpp
rcjsuen: @seemk Thank you for the review and for catching this. I have replaced the `clear()` call with the destructor.

seemk:
Taking a look at the CI issue

instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -178,7 +178,16 @@ TraceContext* GetTraceContext(ngx_http_request_t* req) {
...
+  std::unordered_map<ngx_http_request_t*, TraceContext*>* map = (std::unordered_map<ngx_http_request_t*, TraceContext*>*)val->data;
+  unsigned count = map->count(req);
+  if (count == 1) {
+    return map->at(req);
+  } else if (count > 1) {
+    ngx_log_error(NGX_LOG_ERR, req->connection->log, 0, "More than one TraceContext found for a request");
+    return nullptr;
+  }
+  ngx_log_error(NGX_LOG_ERR, req->connection->log, 0, "TraceContext not found");
+  return nullptr;
```
seemk, line 190:
```suggestion
auto it = map->find(req);
if (it != map->end()) {
  return it->second;
}
ngx_log_error(NGX_LOG_ERR, req->connection->log, 0, "TraceContext not found");
return nullptr;
```

seemk: ^ Just a nitpick as afaik `count` can't be greater than 1 for `unordered_map` and using find is 1 lookup instead of 2 for `count` and `at`

rcjsuen: @seemk You're right. I reread the documentation and that is the case.

seemk (approved):
Thank you, @rcjsuen

Issue #79, opened 2021-12-06
logging trace headers in apache 2.4


jvilhuber:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your platform, build system, version numbers of installed dependencies, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main branch.
Running with apache httpd 2.4.51 (latest at the time of this writing), using mod_proxy to forward requests. Using the mod-otel 0.1.0 from here https://github.com/open-telemetry/opentelemetry-cpp-contrib/releases/download/httpd%2Fv0.1.0/ubuntu-18.04_mod-otel.so.zip

I tried logging the trace header with '%{traceparent}i' but got nothing (a dash/'-' which in apache-speak means "not found"). I assume this is because the header is added by mod-otel AFTER the logging tried to capture it.

**Steps to reproduce**
Using one of the example opentelemetry.conf configs (let's say using 'file' and letting traces go to stderr, i.e. error.log), make a request to my echoserver, which is behind a ProxyPass:

    ProxyPass "/echoserver" "http://echoserver:8080" enablereuse=off

I see the headers coming back from the echo server:

    traceparent=00-e2cb9baa8199a41e746be690a48d6da2-3e2a317ff7539e46-01
    tracestate=

Adding to the LogFormat:

    traceparent=%{traceparent}i

Gets me:

    10.32.3.1 [2021-12-06T11:22:09.329888Z] "GET /echoserver HTTP/1.1" 200 14009 22ms xff=- ua="Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:94.0) Gecko/20100101 Firefox/94.0" traceparent=- gw=10.32.3.1 inBytes=2163 OutBytes=14009

I assume this is likely just the way it works and not a bug. If anyone has suggestions how to get the traceID logged, that would be fantastic.

**What is the expected behavior?**
I was hoping to see the traceparent header in the logs.

**What is the actual behavior?**
A dash for the header (which in apache means it was not found).

**Additional context**
Nothing comes to mind.

TomRoSystems:
Well your thinking is correct I was able to log `traceparent` with `traceparent=%{traceparent}i` with mod_otel options:
[...]

PR #97, opened 2022-01-05
httpd: Expose trace to environment variables + test

TomRoSystems:
Fix for https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/79

### DIFF
--- a/instrumentation/httpd/src/otel/mod_otel.cpp
+++ b/instrumentation/httpd/src/otel/mod_otel.cpp
@@ -37,6 +37,11 @@ using namespace httpd_otel;
 const char kOpenTelemetryKeyNote[] = "OTEL";
 const char kOpenTelemetryKeyOutboundNote[] = "OTEL_PROXY";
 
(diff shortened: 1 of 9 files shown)


instrumentation/httpd/src/otel/mod_otel.cpp
```diff
@@ -99,6 +104,28 @@ static void opentel_child_created(apr_pool_t*, server_rec*)
   initTracer();
 }
 
+// populates environment variables which can be used by other parts of httpd
+void addEnvVars(apr_table_t *envTable, opentelemetry::trace::SpanContext ctx)
+{
+  union {
+    char bfr[33] = {0};
```
seemk, line 111: Is union member initialization defined behaviour?

seemk (approved):

instrumentation/httpd/src/otel/mod_otel.cpp
TomRoSystems: Yes. As it's very similar to `struct`.

PR #98, opened 2022-01-16
Added option to globally ignore paths from beeing traced

tobiasstadler:


seemk (approved):
Thanks! Could you perhaps add an example config line to the README section as well? :slightly_smiling_face:

tobiasstadler:
There is an example in instrumentation/nginx/test/conf/nginx.conf which is linked from the `Usage` section in the README. Should I also add it to the sample nginx.conf in the `Usage` section?

seemk:
> There is an example in instrumentation/nginx/test/conf/nginx.conf which is linked from the `Usage` section in the README. Should I also add it to the sample nginx.conf in the `Usage` section?

Ah no, all good then :+1:

tobiasstadler:
Thank You!

PR #138, opened 2022-04-05
Added a dockerfile for ubuntu20.04 with dependencies of webserver agent

DebajitDas:
More changes to build in ubuntu would be added in other PRs


kpratyus (approved):

rcjsuen (commented):
Would it make more sense to put the `docker` folder inside the `instrumentation/otel-webserver-module` folder? The current path name gives no indication to the user that it is for otel-webserver-module.

DebajitDas:
> Would it make more sense to put the `docker` folder inside the `instrumentation/otel-webserver-module` folder? The current path name gives no indication to the user that it is for otel-webserver-module.

Thanks for pointing this out. It should be insided instrumentation/otel-webserver-module'

lalitb (approved):

PR #151, opened 2022-04-20
Centos7 Suport Dockerfile changes

kpratyus:
Adding Dockerfile changes for Centos7 support

### DIFF
--- a/instrumentation/otel-webserver-module/Dockerfile
+++ b/instrumentation/otel-webserver-module/Dockerfile
@@ -219,19 +219,20 @@ RUN mkdir build-dependencies \
     && mv -f httpd-2.4.23 build-dependencies
 
 # Build Webserver Module
-RUN git clone https://github.com/open-telemetry/opentelemetry-cpp-contrib.git \
-    && cp -r /opentelemetry-cpp-contrib/instrumentation/otel-webserver-module . \
-    && cp -r /dependencies /otel-webserver-module/ \
+COPY . /otel-webserver-module
+
+RUN cp -r /dependencies /otel-webserver-module/ \
     && cp -r /build-dependencies /otel-webserver-module/ \
     && cd otel-webserver-module \
-    && ./gradlew assembleApacheModule
+    && ./gradlew assembleApacheModule -DtargetSystem=centos
 
 # install webserver module
 RUN cd /otel-webserver-module/build \
     && tar -xf opentelemetry-webserver-sdk-x64-linux.tgz \
     && mv -f opentelemetry-webserver-sdk /opt/ \
     && cd ../ \
-    && cp opentelemetry_module.conf /etc/httpd/conf.d/ \
+    && cp opentelemetry_module.conf /etc/httpd/conf/ \
+    && echo 'Include conf/opentelemetry_module.conf' >> /etc/httpd/conf/httpd.conf \
     && cd /opt/opentelemetry-webserver-sdk \
     && ./install.sh \
     && cd /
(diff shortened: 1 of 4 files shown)


instrumentation/otel-webserver-module/test/ApacheTesting.sh
```diff
@@ -66,6 +72,10 @@ if [ $targetSystem = "ubuntu" ]; then
 	targetConfFile='/etc/apache2/opentelemetry_module.conf'
 fi
 
+if [ $targetSystem = "centos7" ]; then
```
DebajitDas, line 75: Should we use elif here instead? This makes it more clearer

instrumentation/otel-webserver-module/test/ApacheTesting.sh
```diff
@@ -75,6 +85,12 @@ if [ $targetSystem = "ubuntu" ]; then
 	fi
 fi
 
+if [ $targetSystem = "centos7" ]; then
```
DebajitDas, line 88: should we use elif here as well?

instrumentation/otel-webserver-module/test/ApacheTesting.sh
```diff
@@ -11,6 +11,12 @@ if [ $targetSystem = "ubuntu" ]; then
 	OTelApacheModule="/opt/opentelemetry-webserver-sdk/WebServerModule/Apache/libmod_apache_otel.so"
 fi
 
+if [ $targetSystem = "centos7" ]; then
```
DebajitDas, line 14: its better if we move this in line no 10 with LOGICAL OR condition, since the value is same in both case.

instrumentation/otel-webserver-module/Dockerfile
```diff
@@ -219,19 +214,20 @@ RUN mkdir build-dependencies \
...
-RUN git clone https://github.com/open-telemetry/opentelemetry-cpp-contrib.git \
-    && cp -r /opentelemetry-cpp-contrib/instrumentation/otel-webserver-module . \
-    && cp -r /dependencies /otel-webserver-module/ \
+COPY . /otel-webserver-module
+
+RUN cp -r /dependencies /otel-webserver-module/ \
     && cp -r /build-dependencies /otel-webserver-module/ \
     && cd otel-webserver-module \
-    && ./gradlew assembleApacheModule
+    && ./gradlew assembleApacheModule -DtargetSystem=centos
```
DebajitDas, line 222: The default value of targetSystem is "centos6".  Since this is default, we no longer need this option here.

DebajitDas (commented):
I also feel we should use ARG VERSION=variable for all versions of dependency. That makes it easier to maintain tomorrow if any version needs to be updated.

kpratyus:
> I also feel we should use ARG VERSION=variable for all versions of dependency. That makes it easier to maintain tomorrow if any version needs to be updated.
> Secondly, for boost we could use wget https://boostorg.jfrog.io/artifactory/main/release/1.75.0/source/boost_1_75_0.tar.gz instead of cloning the boost code. This seems faster. Let me know if you think otherwise on this.

Yeah we can do that

instrumentation/otel-webserver-module/Dockerfile
kpratyus: Resolved

instrumentation/otel-webserver-module/test/ApacheTesting.sh
kpratyus: Done

instrumentation/otel-webserver-module/test/ApacheTesting.sh
kpratyus: done

DebajitDas (approved):

PR #161, opened 2022-05-12
Https is ignored when configured by env var

ElementalWarrior:
Basically if you configure the endpoint via environment variable, it will always to try communicate via insecure. Because use_ssl_credentials is used in the grpc exporter, but we never set it. And it can be configured via conf file, but only if you don't use the env variable.

### DIFF
--- a/instrumentation/nginx/src/agent_config.cpp
+++ b/instrumentation/nginx/src/agent_config.cpp
@@ -21,6 +21,7 @@ static bool SetupOtlpExporter(toml_table_t* table, ngx_log_t* log, OtelNgxAgentC
   auto endpoint_from_env = std::getenv(otel_exporter_otlp_endpoint_env);
 
   if (endpoint_from_env) {
+    config->exporter.use_ssl_credentials = std::string(endpoint_from_env).find("https://") != std::string::npos;
     config->exporter.endpoint = endpoint_from_env;
     return true;
   }


ElementalWarrior:
@seemk  or @kpratyus or @DebajitDas would someone give a review? I can't be the only one encountering this

instrumentation/nginx/src/agent_config.cpp
```diff
@@ -21,6 +21,7 @@ static bool SetupOtlpExporter(toml_table_t* table, ngx_log_t* log, OtelNgxAgentC
   auto endpoint_from_env = std::getenv(otel_exporter_otlp_endpoint_env);
 
   if (endpoint_from_env) {
+    config->exporter.use_ssl_credentials = std::string(endpoint_from_env).find("https://") != std::string::npos;
```
DebajitDas, line 24: Please correct me if I am wrong. Don't we need to assign ca_cert_path or ca_cert_string in this case?

ElementalWarrior: Only if you want to verify the cert. Which would prevent mitm and server impersonations potentially. This enables you to at least connect to an https endpoint configured by env var... At all.
I'm not looking to become a maintainer. But without this you can't hit endpoints only hosted via https without a config file change.

DebajitDas: Ok, make sense.

DebajitDas (commented):
LGTM

DebajitDas:
Actually, I am not the owner of this directory. So I won't be able to merge

ElementalWarrior:
This has been sitting here for over two years. We stopped using otel on nginx, because it was painful to apply the module to nginx.

And no one seems to care enough to review and merge pull requests.

PR #162, opened 2022-05-16
Add license

kpratyus:
Adding License for nginx Source code

### DIFF
--- a/instrumentation/otel-webserver-module/build.gradle
+++ b/instrumentation/otel-webserver-module/build.gradle
@@ -65,10 +65,6 @@ project.ext {
 
     BOOST_INCLUDE = "${modDepDir}/boost/1.75.0/include"
     BOOST_LIB = "${modDepDir}/boost/1.75.0/lib/libboost_filesystem.a ${modDepDir}/boost/1.75.0/lib/libboost_system.a"
-    if (target_system == "ubuntu") {
-	   BOOST_LIB = '-lboost_filesystem -lboost_system'
-	   BOOST_INCLUDE = ''
-   }
 }
 
 task extractANSDK(type: Copy) {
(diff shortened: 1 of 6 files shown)


instrumentation/otel-webserver-module/build.gradle
```diff
@@ -65,10 +65,6 @@ project.ext {
 
     BOOST_INCLUDE = "${modDepDir}/boost/1.75.0/include"
     BOOST_LIB = "${modDepDir}/boost/1.75.0/lib/libboost_filesystem.a ${modDepDir}/boost/1.75.0/lib/libboost_system.a"
-    if (target_system == "ubuntu") {
```
DebajitDas, line 68: Ideally, we should not have this code. This is not there in main. Seems like some rebase error.

instrumentation/otel-webserver-module/test/NginxTesting.sh
```diff
@@ -0,0 +1,93 @@
...
+
+echo "Starting Nginx Server"
+pkill nginx
+nginx # re-start the server
+
+# run load
+curl -m 5 http://localhost:80
+
+echo "Stopping Nginx Server"
+pkill nginx # stop the server
```
DebajitDas, line 93: Did we miss the file inside core api directory?

DebajitDas (changes requested):

instrumentation/otel-webserver-module/test/NginxTesting.sh
kpratyus: It will be used for testing just like Apachetesting.sh

instrumentation/otel-webserver-module/build.gradle
kpratyus: Yeah there was a rebase error in the last push that's why I had to remove this. The rebase error was due to git

DebajitDas (approved):

PR #163, opened 2022-05-18
Nomenclature changes from Appdynamics to OpenTelemetry

DebajitDas:


instrumentation/otel-webserver-module/include/core/api/opentelemetry_ngx_api.h
```diff
@@ -14,8 +14,8 @@
 * limitations under the License.
 */
 
-#ifndef APPDYNAMICS_NGX_API_H
-#define APPDYNAMICS_NGX_API_H
+#ifndef OPENTELEMETRY_NGX_API_H
```
kpratyus, line 17: Aren't these nomenclatures related to core directory?

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_log.c
```diff
@@ -31,7 +31,7 @@ void ngx_writeTrace(ngx_log_t *log, const char* funcName, const char* format, ..
 
     if (logState && log)
     {
-        ngx_log_error(NGX_LOG_ERR, log, 0, "mod_appdynamics: %s: %s", funcName, note);
```
kpratyus, line 34: mod_'o'penetelemetry would be better?

kpratyus (changes requested):

instrumentation/otel-webserver-module/include/core/api/opentelemetry_ngx_api.h
DebajitDas: This nomenclature changes are done for all nginx code added. Gradually, its better we do the same for all other files in core directory.

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_log.c
DebajitDas: ok will add here

kpratyus (approved):

PR #164, opened 2022-05-24
Set span status according to Semantic Conventions of Http

gnm444:
Issue :  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#Status

### DIFF
--- a/instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
+++ b/instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
@@ -37,7 +37,7 @@ enum class StatusCode {
 };
 
 enum class SpanType {
-	INTERNAL,
+  INTERNAL,
   SERVER,
   CLIENT
 };
@@ -53,11 +53,11 @@ public:
         const OtelKeyValueMap& attributes) = 0;
 
 	virtual void AddAttribute(const std::string& key,
-		const SpanAttributeValue& value) = 0;
+	const SpanAttributeValue& value) = 0;
 
-  virtual void SetStatus(const StatusCode status, const std::string& desc = "") = 0;
+        virtual void SetStatus(const StatusCode status, const std::string& desc = "") = 0;
 
-  virtual SpanType GetSpanKind() = 0;
+        virtual SpanType GetSpanKind() = 0;
 };
 
 } //sdkwrapper
--- a/instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
+++ b/instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
@@ -21,6 +21,7 @@
 #include <unordered_map>
 #include <chrono>
 
+
 namespace appd {
 namespace core {
 namespace sdkwrapper {
@@ -31,7 +32,14 @@ using namespace opentelemetry;
 
 enum class StatusCode {
   Ok,
-  Error
+  Error,
+  Unset
+};
+
+enum class SpanType {
+	INTERNAL,
+  SERVER,
+  CLIENT
 };
 
 class IScopedSpan {
@@ -48,6 +56,8 @@ public:
 		const SpanAttributeValue& value) = 0;
 
   virtual void SetStatus(const StatusCode status, const std::string& desc = "") = 0;
+
+  virtual SpanType GetSpanKind() = 0;
 };
 
 } //sdkwrapper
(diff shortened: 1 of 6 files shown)


instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
```diff
@@ -31,7 +32,14 @@ using namespace opentelemetry;
...
-  Error
+  Error,
+  Unset
+};
+
+enum class SpanType {
+  INTERNAL,
+  SERVER,
+  CLIENT
 };
```
DebajitDas, line 43: We already have SpanKind defined in ISdkWrapper. Any Challenge in re-using that?

instrumentation/otel-webserver-module/src/core/api/RequestProcessingEngine.cpp
```diff
@@ -19,17 +19,23 @@
...
 #include <unordered_map>
+#include <sstream>
 
 
 namespace appd {
 namespace core {
 
+#define HTTP_ERROR_1XX 100
+#define HTTP_ERROR_4XX 400
+#define HTTP_ERROR_5XX 500
```
DebajitDas, line 37: Just a discussion, Is it not better to use constexpr instead ?

DebajitDas (changes requested):

instrumentation/otel-webserver-module/src/core/api/RequestProcessingEngine.cpp
DebajitDas: or we can use constexpr https://en.cppreference.com/w/cpp/language/constexpr as well.

instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
gnm444: Since they are in ISdkWrapper.h, including that in this header file is leading to circular inclusions. So avoided it.

kpratyus: ISdkWrapper is the parent right? Why would there be a problem of circular inclusion?

instrumentation/otel-webserver-module/src/core/api/RequestProcessingEngine.cpp
kpratyus: 3 defines look a little clumsy. Should we use a different data structure(like a struct)?

gnm444: Addressed this by using constexpr in SdkConstants.h

instrumentation/otel-webserver-module/include/core/sdkwrapper/IScopedSpan.h
gnm444: Moved them to a new file to avoid circular inclusion

instrumentation/otel-webserver-module/include/core/sdkwrapper/SdkEnums.h
```diff
@@ -0,0 +1,37 @@
...
+
+enum class SpanKind {
+  INTERNAL,
+  SERVER,
+  CLIENT
+};
+
+} // sdkwrapper
+} // core
+} // appd
```
DebajitDas, line 37: Add a newline

instrumentation/otel-webserver-module/src/core/api/RequestProcessingEngine.cpp
```diff
@@ -105,29 +104,31 @@ APPD_SDK_STATUS_CODE RequestProcessingEngine::endRequest(
 
     // check for error and set attribute in the scopedSpan.
     if (error) {
+        // Todo : Error is currently coming like eg HTTP Error code:403
+        // Please change it to only erro code
         std::stringstream strValue;
         strValue << error;
 
         unsigned int errorValue;
         strValue >> errorValue;
```
DebajitDas, line 114: Also incorporate Apache changes

instrumentation/otel-webserver-module/include/core/sdkwrapper/SdkEnums.h
gnm444: Done

instrumentation/otel-webserver-module/src/core/api/RequestProcessingEngine.cpp
gnm444: Done

DebajitDas (approved):
LGTM

DebajitDas:
LGTM, will merge once Pratyush approves as well

kpratyus (approved):

PR #166, opened 2022-05-30
Document inheritance rules of opentelemetry_propagate

Symbianx:
While implementing nginx propagation on our services we noticed that the same inheritance rules of `proxy_set_header` apply to the `opentelemetry_propagate` configuration.

This PR documents that behaviour.


seemk:
Thanks!

Issue #143, opened 2022-04-12
Nginx auth_request does not share parent trace id


phillipuniverse:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your platform, build system, version numbers of installed dependencies, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main branch.

I am in the process of converting to OpenTracing to OpenTelemetry for Nginx. Most of this went ok, but it appears that sub requests from `auth_request` has their own trace_id. I would expect that the trace_id would be the same for the parent and sub request with just different span ids. Below are the files I am using to reproduce the issue.

**Steps to reproduce**

`Dockerfile`:

```
FROM nginx:1.21.6

RUN rm /etc/nginx/conf.d/default.conf

# Built from the package at https://github.com/nginxinc/nginx-unsupported-modules
COPY --from=ghcr.io/nginxinc/amd64/ngx_otel_module:linux-libc-nginx-1.21.6 /usr/lib/nginx/modules/otel_ngx_module.so /usr/lib/nginx/modules/otel_ngx_module.so

ADD localdev.conf /etc/nginx
ADD otel-config.toml /etc/nginx

CMD ["nginx", "-c", "/etc/nginx/localdev.conf", "-g", "daemon off;"]
```

`localdev.conf`:

```
# Available environment variables from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md
env OTEL_EXPORTER_OTLP_ENDPOINT;
env OTEL_RESOURCE_ATTRIBUTES;

load_module /usr/lib/nginx/modules/otel_ngx_module.so;

worker_rlimit_nofile 20000;  # max usable file descriptors, should be about double the worker_connections and less than the OS `nofile` ulimit

events {
    worker_connections 10000;
}

http {
    include /etc/nginx/mime.types;

    # Configuration options at https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/instrumentation/nginx
    # Other examples at https://github.com/nginxinc/bank-of-sirius/tree/master/src/nginx

    opentelemetry_config /etc/nginx/otel-config.toml;

    opentelemetry on;
    opentelemetry_operation_name "$request_method $uri";
    opentelemetry_trust_incoming_spans off;  # Nothing upstream should give us a trace/span
    opentelemetry_attribute http.user_agent $http_user_agent;
[...]

PR #174, opened 2022-06-03
Handling span creation/end on Nginx internal redirection and named location.

DebajitDas:
Issue: During internal redirection, which happens when url is "/",  root span is not visible in the backend.
Root Cause: During internal redirection, module ctx handler array is freed up by nginx code. Due to this, root span created at the start of the request is not ended in otel code. This internal redirection is handled by index, random_index and try_files module. 
Fix: In such modules special handling is done to store the ctx and re-assign once new internal request is created.

The above issue occurs in case of named location configuration also. this is handled by try_files module. Extra special handling needs to be done at this case as well.

There is also a scenario of subrequests, which we might have confirm whether the issue occurs there as well.


phillipuniverse:
@DebajitDas this seems like it would fix #143, what do you think?

DebajitDas:
> 

It does not. This code will on different instrumentation folder "otel-webserver-module" and the one you are referring to is on "nginx" folder.

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -729,40 +751,56 @@ static void otel_payload_decorator(ngx_http_request_t* r, APPD_SDK_ENV_RECORD* p
 /*
     Stopping an Interaction
 */
-static void otel_stopInteraction(ngx_http_request_t* r, const char* module_name)
+static void otel_stopInteraction(ngx_http_request_t* r, const char* module_name,
+    const char* request_handle_key)
```
kpratyus, line 755: It's better to keep request_handle_key as void* ig

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -965,26 +1003,36 @@ static ngx_flag_t ngx_initialize_opentelemetry(ngx_http_request_t *r)
     return false;
 }
 
-static void stopMonitoringRequest(ngx_http_request_t* r)
+static void stopMonitoringRequest(ngx_http_request_t* r,
+    const char* request_handle_key)
 {
-    if(r->internal)
-    {
-        return;
```
kpratyus, line 972: Why do we need to remove this condition?

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -1109,23 +1157,23 @@ static void startMonitoringRequest(ngx_http_request_t* r){
...
+    otel_stopInteraction(r, "ngx_http_limit_conn_module", NULL);
 
     return rvalue;
 }
 
 static ngx_int_t ngx_http_otel_limit_req_handler(ngx_http_request_t *r){
     otel_startInteraction(r, "ngx_http_limit_req_module");
     ngx_int_t rvalue = h[2](r);
-    otel_stopInteraction(r, "ngx_http_limit_req_module");
+    otel_stopInteraction(r, "ngx_http_limit_req_module", NULL);
```
kpratyus, line 1176: the NULL here means that the code should pick the ctx from pool right?

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
DebajitDas: Agreed with void *. But I see anomalies in the code. request handle key inside ngx_http_otel_handles_t is kept as const char*. Hence when we retrieve from ctx, we get const char*. So to keep consistency inside the function, it is kept  as const char*.  Should we change to void* everywhere to be consistent? what are your thoughts?

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
DebajitDas: StopMonitoringRequest would be called from log handler, which comes only once in the lifetime of the request and r->internal can be either 1 or 0. We process only when request handle key is valid.

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -1141,79 +1189,162 @@ static ngx_int_t ngx_http_otel_realip_handler(ngx_http_request_t *r){
...
+*/
+
 static ngx_int_t ngx_http_otel_index_handler(ngx_http_request_t *r){
+    bool old_internal = r->internal;
+    ngx_http_otel_handles_t* old_ctx;
+
     otel_startInteraction(r, "ngx_http_index_module");
+    old_ctx = ngx_http_get_module_ctx(r, ngx_http_opentelemetry_module);
     ngx_int_t rvalue = h[11](r);
-    otel_stopInteraction(r, "ngx_http_index_module");
```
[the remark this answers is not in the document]
DebajitDas: Are you talking about ngx_http_set_ctx(r, old_ctx, ngx_http_opentelemetry_module)?
This is a good point. What I see in code, ctx->propagationHeaders is only set but never used. This needs to be re-visited or taken up later freshly.

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -1141,79 +1189,162 @@ static ngx_int_t ngx_http_otel_realip_handler(ngx_http_request_t *r){
...
+        otel_stopInteraction(r, "ngx_http_index_module", NULL);
+    }
 
     return rvalue;
 }
 
 static ngx_int_t ngx_http_otel_random_index_handler(ngx_http_request_t *r){
+    bool old_internal = r->internal;
+    ngx_http_otel_handles_t* old_ctx;
+
```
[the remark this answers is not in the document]
DebajitDas: Are you saying this can done similar to try_files handler? If that is the case, yes it can done, will do it.

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -729,40 +751,56 @@ static void otel_payload_decorator(ngx_http_request_t* r, APPD_SDK_ENV_RECORD* p
...
-            strcpy(msg, status);
-            strcat(msg, code);
-        }
-        ngx_writeTrace(r->connection->log, __func__, "Stopping the Interaction for: %s", module_name);
-        APPD_SDK_STATUS_CODE res = stopModuleInteraction(ctx->otel_req_handle_key, backendName, backendType, errCode, msg);
-        if (APPD_ISFAIL(res))
-        {
-            ngx_writeError(r->connection->log, __func__, "Error: Stop Interaction failed, result code: %d", res);
-        }
+        otel_req_handle_key = request_handle_key;
```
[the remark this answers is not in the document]
DebajitDas: In this case pointer is copied and not the actual object. Actual object gets deleted in endRequest of RequestProcessingEngine function. So, request_handle_key going out of scope would not affect this.

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -1141,79 +1189,162 @@ static ngx_int_t ngx_http_otel_realip_handler(ngx_http_request_t *r){
 
     otel_startInteraction(r, "ngx_http_realip_module");
     ngx_int_t rvalue = h[3](r);
```
gnm444, line 1191: If possible the logical explanation of hard code values like 2 or 3 can help code readability

DebajitDas: Makes sense, will add this as well

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
DebajitDas: Done

instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
DebajitDas: Done

kpratyus (approved):

DebajitDas:
> @kpratyus @DebajitDas is there any sort of doc or discussion or anything about the diffferences between what you guys have been working on in this `otel-webserver-module` vs the other instrumentation in `instrumentation/httpd` and `instrumentation/nginx`?

@phillipuniverse Thanks for asking this question. Initially, before contributing our source code to OpenSource, we had a discussion in C++ SIG meeting about 'otel-webserver-module' and how is it different from the 'instrumentation/httpd'. We shared the following documents https://docs.google.com/document/d/1yCeeWXAgT9rtXsy3InLlLkHDzSG7zjNNRCvo8_p6wmE/edit
https://docs.google.com/presentation/d/1n4K9M0iSl_I0D7mvfmM6XWszwbFf4BDhh7s6Uagnysk/edit#slide=id.g10eb13f79f8_0_6
These documents demonstrates the architecture and features of Apache only. But same concept applies to nginx as well.
Let me know if you are able to access this document.

phillipuniverse:
@DebajitDas thanks for that. I can't see the document but I did ask for view permissions.

DebajitDas:
> @DebajitDas thanks for that. I can't see the document but I did ask for view permissions.

@phillipuniverse I have provided you the access to those documents. You should be able to view the documents now. Additionally, you can also refer https://opentelemetry.io/blog/2022/instrument-apache-httpd-server/ related to Apache Server.
We are yet to publish anything related to Nginx. So, you would not find any details related to Nginx but overall idea remains same.

PR #177, opened 2022-06-09
Update build-matrix to include ubuntu 22.04

cedricziel:


lalitb:
@cedricziel  - It looks like CI is failing. Seems gRPC/boringssl build is breaking with the newer gcc compiler. otel-cpp is still not tested on the ubuntu 22.04. Looks like gRPC needs to be upgraded. 
fyi -  @tobiasstadler is approver for Nginx.

ThomsonTan:
> @cedricziel - It looks like CI is failing. Seems gRPC/boringssl build is breaking with the newer gcc compiler. otel-cpp is still not tested on the ubuntu 22.04. Looks like gRPC needs to be upgraded. fyi - @tobiasstadler is approver for Nginx.

Could we support ubuntu 22.04 in otel-cpp for now?

owent:
There are some warnings that will break the compiling when using new gcc and clang.I have some patches for the compiling problems in another project, it may be helpful.

+ BoringSSL: https://github.com/atframework/cmake-toolset/blob/main/ports/ssl/boringssl/boringssl-4fb158925f7753d80fb858cb0239dff893ef9f15.patch
+ gRPC 1.46.2: https://github.com/atframework/cmake-toolset/blob/main/ports/grpc/grpc-v1.46.2.patch
+ gRPC 1.43.2: https://github.com/atframework/cmake-toolset/blob/main/ports/grpc/grpc-v1.43.2.patch

PR #183, opened 2022-06-30
Address spec compliance issue (SetName method is deprecated and removed)

maxgolov:
`SetName` method is now removed in the mainline base class, based on Log spec update here:
https://github.com/open-telemetry/opentelemetry-cpp/pull/1383
https://github.com/open-telemetry/opentelemetry-cpp/commit/280f546e175070b324efc966064a634d300bdf23

This causes the contrib `fluentd` build break when the code is compiled with latest OpenTelemetry SDK!

List of changes:

- **Code change** : remove `override` attribute in fluentd `log/recordable.h` . But keep the method for now. Need to figure out what to do with it in the long run. One possibility is to shim the method to `SetAttribute("name", value)` in a way that won't be affecting the code / customers that might still be using it.

- **Code change** : resolve compilation warnings in examples if `HAVE_CONSOLE_LOG` has been defined as a build option in the upper level makefile.

- **NEW - non-invasive build infra option** : allow to build `fluentd` exporter inside the main build tree via `add_directory`. Main idea here is borrowed from nlohmann-json project `CMakeLists.txt`. If fluentd exporter is plugged into the build via `add_directory` into main OpenTelemetry C++ SDK, then don't really need to post a separate package of it. And don't need to provide own config. As a result - artifacts get deployed as part of the main project install, in one build loop rather than in two build loops. Much faster. The other option (clone and build as main project) still there, without any functional changes to its behavior. I'll supply a separate `README.md` for that "fast deployment" later. That's what we use in our Docker image builds / CI pipeline elsewhere.

### DIFF
--- a/exporters/fluentd/CMakeLists.txt
+++ b/exporters/fluentd/CMakeLists.txt
@@ -14,20 +14,28 @@
 
 cmake_minimum_required(VERSION 3.12)
 
-project(opentelemetry-fluentd)
-
+# MAIN_PROJECT CHECK
+## determine if fluentd exporter is built as a subproject (using add_subdirectory) or if it is the main project
+##
+set(MAIN_PROJECT OFF)
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  project(opentelemetry-fluentd)
+  set(MAIN_PROJECT ON)
+endif()
 add_definitions(-DHAVE_CONSOLE_LOG)
 add_definitions(-DENABLE_LOGS_PREVIEW)
 
-find_package(opentelemetry-cpp QUIET)
-if(opentelemetry_cpp_FOUND)
-  message("Using external opentelemetry-cpp")
-else()
-  include(cmake/opentelemetry-cpp.cmake)
-  build_opentelemetry()
-  set(OPENTELEMETRY_CPP_INCLUDE_DIRS "")
-  set(OPENTELEMETRY_CPP_LIBRARIES "opentelemetry::libopentelemetry")
-  message("\nopentelemetry-cpp package was not found. Cloned from github")
+if (MAIN_PROJECT)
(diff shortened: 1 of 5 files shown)


lalitb:
Thanks for the PR. While we are doing this, should we also change the otel-cpp version used internally (if not already installed) here - 

https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/42d82523e162ee1adacaf9e6485926063fa8900e/exporters/fluentd/cmake/opentelemetry-cpp.cmake#L2

Else, the build may fail.

exporters/fluentd/CMakeLists.txt
```diff
@@ -14,20 +14,28 @@
 
 cmake_minimum_required(VERSION 3.12)
 
-project(opentelemetry-fluentd)
-
+# MAIN_PROJECT CHECK
+## determine if fluentd exporter is built as a subproject (using add_subdirectory) or if it is the main project
+##
+set(MAIN_PROJECT OFF)
```
lalitb, line 20: I understand, with `add_subdirectory` added inside the otel-cpp cmake file, it would get build just like any other exporter inside otel-cpp repo. And we don't have that support in otel-cpp ?

maxgolov:
> Else, the build may fail.

Done. Updated to latest.

lalitb (approved):

exporters/fluentd/CMakeLists.txt
maxgolov: In fact, we don't. However, elsewhere when we build a consolidated exporter for [Linux mdsd](https://github.com/Azure/fluentd-plugin-mdsd) -- we end up modifying the main project `CMakeLists.txt` anyways. What I'd like to show is how to add a 1-liner to build the main project _WITH_ fluentd exporter, in essence.

There is no change to current build. You can still build it as a standalone project as before. The new option is additive, not altering existing build behavior.

PR #185, opened 2022-07-07
Adding support for Propagation Headers

kpratyus:
This PR fixes the bug which allows Downstream child span to be part of Parent Span.

### DIFF
--- a/instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
+++ b/instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
@@ -686,6 +686,13 @@ static APPD_SDK_STATUS_CODE otel_startInteraction(ngx_http_request_t* r, const c
         {
             ngx_writeError(r->connection->log, __func__, "Error: Interaction begin result code: %d", res);
         }
+        for(int i=0;i<ix;i++)
+        {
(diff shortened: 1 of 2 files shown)


instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c
```diff
@@ -686,6 +686,13 @@ static APPD_SDK_STATUS_CODE otel_startInteraction(ngx_http_request_t* r, const c
         {
             ngx_writeError(r->connection->log, __func__, "Error: Interaction begin result code: %d", res);
         }
+        for(int i=0;i<ix;i++)
+        {
+          if(propagationHeaders[i].name)
+            free(propagationHeaders[i].name);
```
DebajitDas, line 692: This seems dangerous. 
In line 712,  "h->key.data = propagationHeaders[i].name;"  ---> here h->key.data points to the memory address that  propagationHeaders[i].name; is referring to. Therefore, in 692 when we free up the memory, h->key.data becomes dangling.
Suggestion is to do deep copy (allocate a memory in nginx memory pool) in line 712 and 719.

DebajitDas (changes requested):

DebajitDas (approved):
lgtm

PR #222, opened 2022-10-18
Remove unnecessary baggage and unwanted headers from nginx and apache

DebajitDas:


instrumentation/otel-webserver-module/include/core/api/opentelemetry_ngx_api.h
```diff
@@ -25,11 +25,8 @@
 extern "C" {
 #endif
 
-#define BAGGAGE "baggage"
-#define TRACEPARENT "traceparent"
-#define TRACESTATE "tracestate"
-
-const char* httpHeaders[] = {"baggage", "traceparent", "tracestate"};
+const char* httpHeaders[] = {"traceparent", "tracestate"};
+const size_t headers_len = sizeof(httpHeaders)/sizeof(httpHeaders[0]);
```
kpratyus, line 29: Size of httpdHeaders[0] and httpdHeaders[1] maybe different right? headers_len may not be integer

kpratyus (approved):

PR #261, opened 2023-02-17
Link should point to OT Repository

SaschaBrechmannVHV:
Link should point to OT Repository and not to Cisco Repository


DebajitDas (approved):
Thanks for making this change. LGTM

PR #264, opened 2023-03-03
[Geneva Metrics Exporter]  - Add support for abstract socket

lalitb:
Add support for abstract socket. From [man-page:](https://man7.org/linux/man-pages/man7/unix.7.html)

```
   *  abstract: an abstract socket address is distinguished (from a
          pathname socket) by the fact that sun_path[0] is a null byte
          ('\0').  The socket's address in this namespace is given by
          the additional bytes in sun_path that are covered by the
          specified length of the address structure.  (Null bytes in the
          name have no special significance.)  The name has no
          connection with filesystem pathnames.  When the address of an
          abstract socket is returned, the returned addrlen is greater
          than sizeof(sa_family_t) (i.e., greater than 2), and the name
          of the socket is contained in the first (addrlen -
          sizeof(sa_family_t)) bytes of sun_path.
```

The changes are
 - Remove dependency of [URLParser](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/include/opentelemetry/ext/http/common/url_parser.h) to parse the connection string for UDS, Abstract socket and ETW. As URLParser is not able to parse the abstract socket, and parsing logic is very straightforward for these
 transport mechanisms.
- Add unit-test case to test abstract sockets.
- Add stress test for Linux.


exporters/geneva/CMakeLists.txt
```diff
@@ -95,6 +95,9 @@ if(BUILD_EXAMPLE)
   add_executable(example_metrics example/example_metrics.cc
                                  example/foo_library.cc)
   target_link_libraries(example_metrics opentelemetry_exporter_geneva_metrics)
+
+  add_executable(stress_test_linux example/stress_test_linux.cc)
+target_link_libraries(stress_test_linux opentelemetry_exporter_geneva_metrics)
```
ThomsonTan, line 100: nit: align this with the previous line.

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
```diff
@@ -252,6 +254,12 @@ struct SocketAddr {
       // Max length of Unix domain filename is up to 108 chars
       strncpy_s(m_data_un.sun_path, sizeof(m_data_un.sun_path),
                 unix_domain_path, sizeof(m_data_un.sun_path));
+      // special handling for abstract socket - they should be
+      // prefixed with '@' character.
+      if (unix_domain_path[0] == '@') {
+        m_data_un.sun_path[0] = '\0';
```
ThomsonTan, line 260: Why do we need set `sun_path[0]` to NULL?

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
```diff
@@ -306,7 +314,9 @@ struct SocketAddr {
   size_t size() const {
 #ifdef HAVE_UNIX_DOMAIN
     // Unix domain struct m_data_un
-    if (isUnixDomain)
+    if (isUnixDomain && abstract_socket_size)
```
ThomsonTan, line 317: Note that checking for `isUnixDomain` may be unnecessary in this context since it can be inferred that `abstract_socket_zie > 0` implies the use of Unix Domain Sockets (UDS).

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
```diff
@@ -306,7 +314,9 @@ struct SocketAddr {
   size_t size() const {
 #ifdef HAVE_UNIX_DOMAIN
     // Unix domain struct m_data_un
```
ThomsonTan, line 316: Move this comment to the else if statement below?

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
lalitb: The abstract socket starts with a NULL char as the first character. We ask for the socket name in the exporter configuration and prefix it with '@' (the printable char for abstract socket) before passing it to socket_tools.h. And then replace '@' with '\0' here.

https://man7.org/linux/man-pages/man7/unix.7.html

```
abstract: an abstract socket address is distinguished (from a
          pathname socket) by the fact that sun_path[0] is a null byte
          ('\0'). 
```

NULL char in `char *` can mess with the string processing, so it is replaced in the end.

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
lalitb: yes good point. Having both just makes it bit more readable, so kept it :)

exporters/geneva/CMakeLists.txt
lalitb: Thanks this is fixed now.

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
lalitb: Thanks this is fixed now.

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
ThomsonTan: Why is it necessary to use the `@` symbol to represent `NULL` character in the exporter configuration, rather than assuming that it will be automatically prepended to the user's input path? Additionally, must we continue to support non-abstract UDS?

lalitb: So, these are the connection strings we support as export options:

   `Account={MetricAccount};NameSpace={MetricNamespace}`     - ETW
  `Endpoint=unix://{UDS Path};Account={MetricAccount};Namespace={MetricNamespace}`   - UDS (abstract/non-abstract)
If `{UDS path}` is prefixed with `@` character, this would mean the USD socket is of abstract type. The UDS non-abstract file path can never have this prefix, so it should be safe to do so.

The other option without prefix `@` character would be:

   `Account={MetricAccount};NameSpace={MetricNamespace}`     - ETW
  `Endpoint=unix://{UDS Path};Account={MetricAccount};Namespace={MetricNamespace}`   - UDS non-abstract
  `Endpoint=abs://{UDS Path};Account={MetricAccount};Namespace={MetricNamespace}`  - UDS abstract
And then pass the abstract flag to socket_tools.h which will create the appropriate socket based on this flag.

It was just the implementation choice :)

Also on Linux, `lsof` and `netstat` commands will display an abstract socket prefixed with this character, so the character is familiar one and logically makes sense:

```bash
$ lsof -U

upowerd   1604       root    5u  unix 0xffff88005af5f400      0t0 18631 type=STREAM
colord    1614     colord   10u  unix 0xffff880034d3f400      0t0 18170 type=STREAM
systemd   2009       root   13u  unix 0xffff88005a293000      0t0 21213 /run/user/0/systemd/notify type=DGRAM
systemd   2009       root   14u  unix 0xffff88005a293c00      0t0 21214 /run/user/0/systemd/private type=STREAM
upstart   1525    lightdm    7u  unix 0xffff880034b99800      0t0 17301 @/com/ubuntu/upstart-session/111/1525 type=STREAM

```

ThomsonTan (approved):

PR #277, opened 2023-04-27
Support Apache Hook translate_name in otel-webserver-module

timvtinsa:
Some Apache http based applications use the **translate_name** hook which is not currently tracked by the otel-webserver module. This pull request offers support for this hook on the same basis as the other Apache httpd hooks supported by the module.


lalitb:
@kpratyus @DebajitDas Can you please help here.

PR #280, opened 2023-05-01
[Geneva Exporter] Add support for exporting metrics to different combinations of account and namespace

lalitb:
- Add support for exporting metrics to different combinations of account and namespace
     - reading and configure account and namespace from the attributes if present.
     - ensuring that the account and namespace is not exported as attributes. They should be used only for configuration.
- Removed an optimization that was made earlier by having separate buffers for histogram and non-histogram metrics with pre-serialized account and namespace. All the metrics share the same buffer now since we have to serialize account and namespace on every export call now.

### DIFF
--- a/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/exporter.h
+++ b/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/exporter.h
@@ -65,10 +65,7 @@ private:
   std::unique_ptr<DataTransport> data_transport_;
 
   // metrics storage
-  char buffer_non_histogram_[kBufferSize];
-  char buffer_histogram_[kBufferSize];
-  uint64_t buffer_index_non_histogram_;
-  uint64_t buffer_index_histogram_;
+  char buffer_[kBufferSize];
 
   size_t InitializeBufferForNonHistogramData();
   size_t InitiaizeBufferForHistogramData();
--- a/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/exporter.h
+++ b/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/exporter.h
@@ -33,6 +33,9 @@ constexpr uint64_t kSecondsToUnixTime =
                   // 1601-01-01T00:00:00Z and UNIX/Linux epoch
                   // (1970-01-01T00:00:00Z)
 
+const std::string kAttributeNamespaceKey = "_microsoft_metrics_namespace";
+const std::string kAttributeAccountKey = "_microsoft_metrics_account";
+
 using ValueType = nostd::variant<int64_t, double>;
 
 /**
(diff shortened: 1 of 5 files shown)


ThomsonTan (commented):
Does the separate buffer cause any issue?

lalitb:
> Does the separate buffer cause any issue?

No as we only do one export at a time, to optimise memory.

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/exporter.h
```diff
@@ -65,13 +68,8 @@ class Exporter final : public opentelemetry::sdk::metrics::PushMetricExporter {
   std::unique_ptr<DataTransport> data_transport_;
 
   // metrics storage
-  char buffer_non_histogram_[kBufferSize];
-  char buffer_histogram_[kBufferSize];
-  uint64_t buffer_index_non_histogram_;
-  uint64_t buffer_index_histogram_;
+  char buffer_[kBufferSize];
```
ThomsonTan, line 71: As the buffer unified, should the buffer size be increased?

lalitb: We always export one metric point at given time, so this buffer will either contain histogram metric point, or non-histogram metric point at given time.  Earlier, we had two buffers as it allowed us to initialize static data (account name, and namespace) for both types of metrics during bootstrapping. As the data is no longer always static, we get rid of two buffers, and only used one.

exporters/geneva/test/metrics_exporter_test.cc
```diff
@@ -23,6 +23,10 @@ std::string kUnixDomainPathAbstractSocket = "@/tmp/ifx_unix_socket";
 const std::string kNamespaceName = "test_ns";
 const std::string kAccountName = "test_account";
 
+const std::string kCustomNamespaceName = "custom_ns";
+const std::string kCustomAccountName = "custom_account";
```
ThomsonTan, line 27: Are these 2 constants used?

lalitb: Good catch. I missed to check-in the latest changes to enable tests for custom namespace and account. Have added them now.

exporters/geneva/test/common/generate_metrics.h
```diff
@@ -94,6 +109,18 @@ GenerateSumDataLongMetrics() {
...
+  opentelemetry::sdk::metrics::PointAttributes attributes = 
+  {{kCounterLongAttributeKey1,
+                kCounterLongAttributeValue1}};
+  if (account_name.size() && account_namespace.size()) {
+    std::cout << "\n Generating custom namespae and metrics for " << kCounterCustomLongValue << "\n";
+    attributes[opentelemetry::exporter::geneva::metrics::kAttributeNamespaceKey] = account_namespace;
+    attributes[opentelemetry::exporter::geneva::metrics::kAttributeAccountKey] = account_name;
+      sum_point_data.value_ = kCounterCustomLongValue;
+  } else  {
+    std::cout << "\n Generating derfault  namespae and metrics for " << kCounterLongValue << "\n";
```
ThomsonTan, line 121:
```suggestion
    std::cout << "\n Generating derfault namespae and metrics for " << kCounterLongValue << "\n";
```

ThomsonTan (approved):

exporters/geneva/test/common/generate_metrics.h
lalitb: Thanks removed the comments.

PR #281, opened 2023-05-07
[WIP] Improve tracing

tiithansen:
1) Add possibility to use HTTP OTLP exporter.
2) Improve tracing so that module creates internal or client span depending if the request is handled by nginx directly or passed upstream. 3) Set span status depending on result. If status code is between 400 - 599 set status as error. 4) Catch internal processing failures in log phase.


tiithansen:
Needs some work

PR #283, opened 2023-05-12
Improve tracing

tiithansen:
1) Add possibility to use HTTP OTLP exporter.
2) Improve tracing so that module creates internal or client span depending if the request is handled by nginx directly or passed upstream. 
3) Set span status depending on result. If status code is between 400 - 599 set status as error.
4) Catch internal processing failures in log phase. 
5) Rewrite tests in JS


tiithansen:
@esigo any plans with this one?

huggsboson:
Just wanted to add an upvote on this one, I reached out to the nginx team and they were resistent to adding client spans to the instrumentation itself, so this seems like the best path forward:
https://github.com/nginxinc/nginx-otel/issues/17

huggsboson:
> I've got mixed feelings regarding span status. For example, it isn't always clear whether 4xx status codes are "errors". Does OTEL specification have clear stance on these?
> 
> Perhaps configuration flag for defining which status codes are viewed as span errors would be enough?

I wouldn't consider 4xx's errors either, every time we did it in our other instrumentation it lead to a lot of false positives / noise.

johanneswuerbach:
@esigo / @seemk what would be required to move this forward?

patsevanton:
is there any progress on this issue?

lalitb:
@seemk Would it be possible for you to review this ?

lalitb:
@tiithansen In case there is no reply here, would you like to own this module? We can add you in owners.txt, and help for this to get merged.

PR #286, opened 2023-05-15
Add OpenTelemetry logs exporter targeting user_events on Linux

ThomsonTan:


lalitb (approved):
LGTM/

exporters/user_events/include/opentelemetry/exporters/user_events/logs/exporter.h
```diff
@@ -0,0 +1,70 @@
...
+
+namespace sdk_logs = opentelemetry::sdk::logs;
+
+/**
+ * The user_events logs exporter exports logs data to Geneva
+ */
+class Exporter final : public opentelemetry::sdk::logs::LogRecordExporter
+{
+public:
+  Exporter(const ExporterOptions &options) noexcept;
```
idigdoug, line 32: To avoid unwanted implicit conversions, it's usually a good idea to make all single-parameter constructors be explicit, e.g. `explicit Exporter(const ExporterOptions& options) noexcept`.

exporters/user_events/include/opentelemetry/exporters/user_events/logs/exporter.h
```diff
@@ -0,0 +1,70 @@
...
+      std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept override;
+
+  bool isShutdown() const noexcept;
+
+private:
+  const ExporterOptions options_;
+  bool is_shutdown_ = false;
+  mutable opentelemetry::common::SpinLockMutex lock_;
+
+  ehd::Provider provider_{"opentelemetry_logs"};
```
idigdoug, line 54: This string is important and will vary based on the logger parameters being used. We need to carefully specify how this string gets determined, as it is very important (affects event filtering) and also has restrictions (can only have ASCII letters, ASCII digits, and underscores).

exporters/user_events/src/recordable.cc
```diff
@@ -0,0 +1,115 @@
...
+namespace user_events
+{
+namespace logs
+{
+
+namespace api_logs = opentelemetry::logs;
+
+Recordable::Recordable() noexcept
+{
+  event_builder_.Reset("OpenTelemetry-Logs");
```
idigdoug, line 22: Isn't the event's name supposed to be used here?

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+                       ehd::EventBuilder &event_builder) noexcept
+{
+  static_assert(nostd::variant_size<api_common::AttributeValue>::value == kAttributeValueSize,
+                "AttributeValue has changed, update PopulateAttributeValue");
+
+  const char *key_name = key.data();
+
+  if (nostd::holds_alternative<bool>(value))
+  {
+    event_builder.AddValue(key_name, nostd::get<bool>(value), event_field_format_default);
```
idigdoug, line 31: event_field_format_boolean

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+  }
+  else if (nostd::holds_alternative<int>(value))
+  {
+    event_builder.AddValue(key_name, nostd::get<int>(value), event_field_format_default);
```
idigdoug, line 35: event_field_format_signed_int

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+  }
+  else if (nostd::holds_alternative<int64_t>(value))
+  {
+    event_builder.AddValue(key_name, nostd::get<int64_t>(value), event_field_format_default);
```
idigdoug, line 39: event_field_format_signed_int

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+  {
+    event_builder.AddValue(key_name, nostd::get<unsigned int>(value), event_field_format_default);
+  }
+  else if (nostd::holds_alternative<uint64_t>(value))
+  {
+    event_builder.AddValue(key_name, nostd::get<uint64_t>(value), event_field_format_default);
+  }
+  else if (nostd::holds_alternative<double>(value), event_field_format_default)
+  {
+    event_builder.AddValue(key_name, nostd::get<double>(value), event_field_format_default);
```
idigdoug, line 51: event_field_format_float

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+                                  event_field_format_default);
+  }
+  else if (nostd::holds_alternative<nostd::string_view>(value))
+  {
+    event_builder.AddString<char>(key_name, nostd::get<nostd::string_view>(value).data(),
+                                  event_field_format_default);
+  }
+  else if (nostd::holds_alternative<nostd::span<const uint8_t>>(value))
+  {
+    // TODO: implement
```
idigdoug, line 65: I think this can be `AddValueRange(key_name, value.begin(), value.end(), event_field_format_default)`

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+void PopulateAttribute(nostd::string_view key,
+                       const api_common::AttributeValue &value,
+                       ehd::EventBuilder &event_builder) noexcept
+{
+  static_assert(nostd::variant_size<api_common::AttributeValue>::value == kAttributeValueSize,
+                "AttributeValue has changed, update PopulateAttributeValue");
+
+  const char *key_name = key.data();
+
+  if (nostd::holds_alternative<bool>(value))
```
idigdoug, line 29: This looks like a std::variant, so maybe there is a `visit` function for it. If there is a `visit` function, it's usually a good idea to use it -- it's usually more efficient than an IF-ELSE chain (it usually compiles into a SWITCH statement), and it helps you make sure you don't accidentally miss any cases.

exporters/user_events/src/utils.cc
```diff
@@ -0,0 +1,87 @@
...
+  {}
+  else if (nostd::holds_alternative<nostd::span<const unsigned int>>(value))
+  {}
+  else if (nostd::holds_alternative<nostd::span<const uint64_t>>(value))
+  {}
+  else if (nostd::holds_alternative<nostd::span<const double>>(value))
+  {}
+  else if (nostd::holds_alternative<nostd::span<const bool>>(value))
+  {}
+  else if (nostd::holds_alternative<nostd::span<const nostd::string_view>>(value))
```
idigdoug, line 79: I think this can be `AddStringRange(...)`

exporters/user_events/include/opentelemetry/exporters/user_events/logs/exporter.h
ThomsonTan: Do you have any suggestion on this? Configure as string literal in the code, so we can make sure it is a valid name for tracepoint, or we could accept string from user and then do a validation? Do we expect the user to configure the provider name here?

exporters/user_events/src/recordable.cc
ThomsonTan: Currently it is hard-coded here. Where do you suggest to move it to?

exporters/user_events/src/utils.cc
ThomsonTan: Thanks for the catch. Just curious, will it make difference to the serialization process with `event_field_format_default`?

idigdoug: Yes. For an 8-bit value, default means "Treat as unsigned integer" (1/0) and bool means "treat as Boolean" (true/false).

exporters/user_events/src/recordable.cc
idigdoug: I think this is supposed to be the OpenTelemetry event name. I believe you get that from SetEventId.

exporters/user_events/include/opentelemetry/exporters/user_events/logs/exporter.h
idigdoug: I think we have to make it configurable on a per-logger basis, and if the user provides an invalid value we need to validate it. Maybe talk to Reiley or Kyle for details - I think they have some ideas about how this should work.

PR #287, opened 2023-05-16
[Geneva Metrics Exporter] Fix stress test for abstract socket.

lalitb:


exporters/geneva/example/stress_test_linux.cc
```diff
@@ -47,7 +47,7 @@ namespace {
   std::cout << " Init metrics : " <<account_name << "-" << ns <<"-"<< socket_path << "-"<< is_uds << "-"<<metrics_collection_time_secs << "\n";
   std::string conn_string = "Account=" + account_name + ";Namespace=" + ns;
   if (!is_uds){
-    std::string null_string(1, '#');
+    std::string null_string(1, '@');
```
ThomsonTan, line 50: Is this a convention in our code, not required by path for abstract UDS, right?

lalitb: Yes, this is the convention in the code, to prefix `@` before abstract socket name.

ThomsonTan (approved):

PR #291, opened 2023-06-03
Initial metrics exporter targeting user_events

ThomsonTan:


exporters/user_events/src/metrics_exporter.cc
```diff
@@ -0,0 +1,109 @@
...
+
+Exporter::Exporter(const ExporterOptions &options)
+: options_(options),
+    aggregation_temporality_selector_{otlp_exporter::OtlpMetricUtils::ChooseTemporalitySelector(options_.aggregation_temporality)}
+{
+  int err;
+
+  err = tracepoint_open_provider(&provider_);
+
+  err = tracepoint_connect(&otlp_metrics_, &provider_, "otlp_metrics");
```
lalitb, line 29: Is "otlp_metrics" the name of tracepoint? In case of multiple exporter instances across applications, do all of them share the same name ?

lalitb (approved):

exporters/user_events/src/metrics_exporter.cc
ThomsonTan: Thanks. Yes. I plan to read this from the config option in later PR. How do you think?

PR #315, opened 2023-08-08
instrumentation/nginx: update documentation to drop deprecated options

mfrw:
opentelemetry-cpp has [deprecated](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CMakeLists.txt#L165) `WITH_OTLP` cmake option.
It suggests using `WITH_OTLP_GRPC/WITH_OTLP_HTTP` instead.

Signed-off-by: Muhammad Falak R Wani <falakreyaz@gmail.com>


marcalff (approved):
Thanks for the fix.

PR #316, opened 2023-08-11
Allow nginx workflow to be triggered from web.

swamidass:
Artifacts expired, so the built shared object files no longer are downloadable. This update allows workflow to be triggered from web to rebuild the files.


swamidass:
Just signed the CLA.

seemk (approved):

PR #329, opened 2023-09-27
Build with otel-cpp v1.11.0 & use Cumulative temporality for UpDownCounter

lalitb:
- Fix the build with otel-cpp v1.11.0
- Use Cumulative Temporality for UpDownCounter.

### DIFF
--- a/exporters/geneva/example/stress_test_linux.cc
+++ b/exporters/geneva/example/stress_test_linux.cc
@@ -28,9 +28,11 @@
 #include <chrono>
 #include <sys/times.h>
 
+#ifdef HAVE_BOOST
 #include <boost/uuid/uuid.hpp>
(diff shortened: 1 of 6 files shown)


exporters/geneva/example/stress_test_linux.cc
```diff
@@ -28,9 +28,11 @@
 #include <chrono>
 #include <sys/times.h>
 
+#ifdef HAVE_BOOST
```
ThomsonTan, line 31: Is boost optional, and missing it doesn't break the build?

lalitb: yeah it's optional. We use some of it's functionality if present, else use that from STL.

ThomsonTan (approved):

PR #333, opened 2023-10-17
Fix mixed readme descriptions

thusoy:
Descriptions were mixed for name and namespace for both apace and nginx modules.


ThomsonTan (approved):
Thanks for the fix.

PR #335, opened 2023-10-25
Fix status mapping in fluentd exporter

ThomsonTan:


exporters/fluentd/src/trace/recordable.cc
```diff
@@ -91,11 +91,9 @@ void Recordable::AddLink(
 
 void Recordable::SetStatus(opentelemetry::trace::StatusCode code,
                            nostd::string_view description) noexcept {
-  if (code != opentelemetry::trace::StatusCode::kUnset) {
-    options_["tags"]["otel.status_code"] = code;
-    if (code == opentelemetry::trace::StatusCode::kError) {
-      options_["tags"]["error"] = description;
-    }
```
lalitb, line 98: Should we keep the these tags, not to break existing queries. Atleast "otel.status_code“ is populated in .Net too.

ThomsonTan: Thanks, reverted the change for "otel.status_code".

lalitb (approved):
Thanks.

PR #357, opened 2023-12-28
[Logs SDK] Log Appender for spdlog

chusitoo:
Implemented as described in https://github.com/open-telemetry/opentelemetry-cpp/issues/2046:

- Sink implemented in spdlog::sinks namespace for commonality with existing sinks.
- Minimal examples provided to demonstrate the setup required.
- Basic unit testing.
- Tried using the same flags as those used in otel-cpp for configuring cmake project.
- Add clang format file to help keep the style in line with otel-cpp repo.


instrumentation/spdlog/src/sink.cc
```diff
@@ -0,0 +1,48 @@
...
+  auto provider   = opentelemetry::logs::Provider::GetLoggerProvider();
+  auto logger     = provider->GetLogger(msg.logger_name.data(), kLibraryName, libraryVersion());
+  auto log_record = logger->CreateLogRecord();
+
+  if (log_record)
+  {
+    using namespace opentelemetry::trace::SemanticConventions;
+
+    log_record->SetSeverity(levelToSeverity(msg.level));
+    log_record->SetBody(opentelemetry::nostd::string_view(msg.payload.data()));
```
ndl, line 31: This should either use `string_view(msg.payload.data(), msg.payload.size())` or `string_view(msg.payload.c_str())` to avoid out-of-bounds access?

instrumentation/spdlog/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+)
+
+if(OPENTELEMETRY_INSTALL)
+  install(
+    TARGETS ${this_target}
+    EXPORT "${PROJECT_NAME}-target"
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
```
ndl, line 37: Probably also makes sense to export the target itself so that it's easier to use this library in another cmake-based project, smth like

```
  install(
    EXPORT "${PROJECT_NAME}-target"
    FILE "${PROJECT_NAME}-target.cmake"
    NAMESPACE "${PROJECT_NAME}::"
    DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})
```
?

Ideally also a config (see e.g. https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#creating-a-package-configuration-file) for easier discoverability / reuse?

instrumentation/spdlog/src/sink.cc
chusitoo: Make sense, thanks! Good catch.

instrumentation/spdlog/CMakeLists.txt
chusitoo: That's reasonable, I'll have a look. I was seeing it as a Makefile-based build only.

marcalff (approved):
Thanks for the contribution.

lalitb (approved):

PR #366, opened 2024-01-24
[Geneva] Fix export name of Geneva metrics exporter

ThomsonTan:
The Geneva metrics exporter is exported as opentelemetry-cpp::geneva which could confuse itself with the metrics SDK.

Also wrap the `install()` call in `OPENTELEMETRY_INSTALL` which is used by the main repo.


lalitb (approved):
Thanks.

PR #408, opened 2024-03-28
Update CMakeLists.txt arguments of Geneva and User Events exporters for Vcpkg build

Pravila00:
# Changes
- Rename `BUILD_EXAMPLE` to `BUILD_EXAMPLES` which seems to be a more standard argument name in CMake for skipping the build of examples. In addition, in [Vcpkg for opentelemetry-cpp port](https://github.com/microsoft/vcpkg/blob/3c76dc55f8bd2b7f4824bcd860055094bfbbb9ea/ports/opentelemetry-cpp/portfile.cmake#L68), `BUILD_EXAMPLES` is the one used.
- Add option to skip the build of Tracepoints library. This will be used for Vcpkg build, as we will set Tracepoint library as a dependency of the feature.

### DIFF
--- a/.github/workflows/geneva_metrics.yml
+++ b/.github/workflows/geneva_metrics.yml
@@ -43,6 +43,6 @@ jobs:
           mkdir -p "$GITHUB_WORKSPACE/otel_cpp_contrib/exporters/geneva/build"
           cd "$GITHUB_WORKSPACE/otel_cpp_contrib/exporters/geneva/build"
           cmake .. "-DCMAKE_PREFIX_PATH=$HOME/prebuilt-otel" \
-            -DBUILD_TESTING=ON -DBUILD_EXAMPLE=ON
+            -DBUILD_TESTING=ON -DBUILD_EXAMPLES=ON
           cmake --build . -j${nproc}
           ctest
--- a/exporters/user_events/CMakeLists.txt
+++ b/exporters/user_events/CMakeLists.txt
@@ -7,8 +7,9 @@ if(WIN32)
   message(FATAL_ERROR "user_events exporter is Linux only for now")
 endif()
 
-option(BUILD_EXAMPLE "Build example" ON)
+option(BUILD_EXAMPLES "Build example" ON)
 option(BUILD_TESTING "Build tests" ON)
+option(BUILD_TRACEPOINTS "Build tracepoints library" ON)
 
 project(opentelemetry-user_events-exporter)
 if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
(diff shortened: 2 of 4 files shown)


exporters/user_events/CMakeLists.txt
```diff
@@ -26,10 +27,12 @@ if(MAIN_PROJECT)
...
 endif()
 
-# don't build samples and tools from LinuxTracepoints.
-set(BUILD_SAMPLES OFF)
-set(BUILD_TOOLS OFF)
-add_subdirectory(third_party/LinuxTracepoints)
+if(BUILD_TRACEPOINTS)
+  # don't build samples and tools from LinuxTracepoints.
+  set(BUILD_SAMPLES OFF)
+  set(BUILD_TOOLS OFF)
```
ThomsonTan, line 33: Restore the potential value in `BUILD_SAMPLES` and `BUILD_TOOLS` after the call to `add_subdirectory()`? This was missed before, but I think it will be great to have it.

Pravila00: I've added the usage of [block()](https://cmake.org/cmake/help/latest/command/block.html) which makes this automatic. Nonetheless, this is added after CMake 3.25, and the CMake file was using 3.12. If updating the CMake version is a blocker, I can reverse this and try a more manual approach storing the values in some variables.

Otherwise, we can keep block() which is a bit cleaner.

ThomsonTan: Seems this version of CMake is not available in the Github CI runner by default.
>  CMake 3.25 or higher is required.  You are running version 3.16.3

Pravila00: Makes sense, I've restored CMake version to 3.12 and just use some variables to store the values before adding the subdirectory. Thanks!

.github/workflows/geneva_metrics.yml
```diff
@@ -43,6 +43,6 @@ jobs:
           mkdir -p "$GITHUB_WORKSPACE/otel_cpp_contrib/exporters/geneva/build"
           cd "$GITHUB_WORKSPACE/otel_cpp_contrib/exporters/geneva/build"
           cmake .. "-DCMAKE_PREFIX_PATH=$HOME/prebuilt-otel" \
-            -DBUILD_TESTING=ON -DBUILD_EXAMPLE=ON
-          cmake --build . -j$(nproc)
+            -DBUILD_TESTING=ON -DWITH_EXAMPLES=ON
+          cmake --build . -j${nproc}
```
ThomsonTan, line 47: This could be a bad merge, here it should be `-j$(nproc)`.

lalitb (approved):

Pravila00:
I've just modified my commit history for using only this GitHub account where I have CLA access, we should be good now :)

PR #413, opened 2024-03-29
[Geneva] Add CI for user_events exporter

ThomsonTan:


.github/workflows/user_events.yml
```diff
@@ -0,0 +1,55 @@
+name: user_events 
+
+on:
+  push:
+    branches:
+      - '*'
```
lalitb, line 6: trigger on `push` only to the `main` branch, and not all?

ThomsonTan: There could be potentially some release branch which should also run the CI? And this condition is copied from below.

https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/4ff2eb2f0b6617dee1ef65091b842e1e1cdcd776/.github/workflows/geneva_metrics.yml#L5

lalitb (approved):
with nit suggestion.

PR #417, opened 2024-04-03
[CI] shellcheck CI step and various shell fixes

johnbley:
- Use `set -e` for almost all shell scripts (one exception which is commented)
- Pass `shellcheck --severity=error` 
- Add CI step to run `shellcheck`

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,19 @@
+name: Shellcheck
+
+on:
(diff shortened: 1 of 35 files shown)


johnbley:
Blergh.  Apparently there is some combination of (a) things that lack of `set -e` was masking which were failing before, and (b) things where I should have been more careful applying `set -e` and that are expected to fail.  Will dig deeper when I have more time.

PR #425, opened 2024-04-06
[CI] Add CI for instrumentation/glog

### LINKED ISSUE #421: [CI] Add CI for instrumentation/glog
Add CI for instrumentation/glog

marcalff:
Fixes #421

Fixed CMakeLists:

* Require GTest 1.12.0. Building with 1.11 breaks the build due to a gtest bug.

Upgraded to google/glog 0.7.0

* Fixed CMakeLists
* Removed `GLOG_VERSION_HAS_LOGMESSAGETIME`
* Fixed warnings related to timstamps.

### DIFF
--- a/.github/workflows/glog.yml
+++ b/.github/workflows/glog.yml
@@ -68,10 +68,10 @@ jobs:
           cmake --build . -j$(nproc)
           cmake --install . --prefix="${GITHUB_WORKSPACE}/sandbox"
 
-      - name: build glog contrib
+      - name: build instrumentation/glog contrib
         run: |
-          mkdir -p "${GITHUB_WORKSPACE}/glog/build"
-          cd "${GITHUB_WORKSPACE}/glog/build"
+          mkdir -p "${GITHUB_WORKSPACE}/instrumentation-glog/build"
+          cd "${GITHUB_WORKSPACE}/instrumentation-glog/build"
           cmake ../../opentelemetry-cpp-contrib/instrumentation/glog \
             -G Ninja \
             -DCMAKE_PREFIX_PATH="${GITHUB_WORKSPACE}/sandbox" \
--- a/.github/workflows/glog.yml
+++ b/.github/workflows/glog.yml
@@ -29,7 +29,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: "google/glog"
-          ref: "v0.7.0"
+          # TODO: upgrade to v0.7.0 and fix deprecation warnings
+          ref: "v0.6.0"
           path: "glog"
           submodules: "recursive"
       - name: checkout opentelemetry-cpp-contrib
--- a/.github/workflows/glog.yml
+++ b/.github/workflows/glog.yml
@@ -28,8 +28,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: "google/glog"
-          # TODO: upgrade to v0.7.0 and fix deprecation warnings
-          ref: "v0.6.0"
+          ref: "v0.7.0"
           path: "glog"
       - name: checkout opentelemetry-cpp-contrib
         uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


marcalff:
@chusitoo FYI

.github/workflows/glog.yml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          repository: "google/googletest"
+          ref: "release-1.12.1"
+          path: "googletest"
+          submodules: "recursive"
+      - name: checkout glog
+        uses: actions/checkout@v3
+        with:
+          repository: "google/glog"
+          # TODO: upgrade to v0.7.0 and fix deprecation warnings
```
chusitoo, line 32: One problem with glog is that it appears to be unversioned and the only way to figure out when something is changed in the API has been by basing it on a specific define that appears when the breakage happens, like [this](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/glog/include/opentelemetry/instrumentation/glog/sink.h#L15C1-L20C7)

Possibily, something similar would need to be done for the deprecation notice around `.timestamp()` too

instrumentation/glog/CMakeLists.txt
```diff
@@ -69,6 +69,8 @@ if(OPENTELEMETRY_INSTALL)
 endif() # OPENTELEMETRY_INSTALL
 
 if(BUILD_TESTING)
```
chusitoo, line 71: To make this available for ctest in CI

```suggestion
if(BUILD_TESTING)
  enable_testing()
```

marcalff: Thanks, fixed.

.github/workflows/glog.yml
marcalff: I used 0.6.0 because of the timestamp deprecation warnings in 0.7.0

This PR is just to add CI, trying to limit the changes to a minimum.

After this, I think we can upgrade to 0.7.0, change the timestamp code to resolve warnings, and require 0.7.0 as a minimum then, also resolving the ifdef around GLOG_VERSION_HAS_LOGMESSAGETIME.

Looking at the glog source, the deprecated methods are removed already from version 0.8.0 to come, so this must be fixed eventually.

.github/workflows/glog.yml
```diff
@@ -0,0 +1,102 @@
...
+    name: CMake Linux
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout googletest
+        uses: actions/checkout@v3
+        with:
+          repository: "google/googletest"
+          ref: "release-1.12.1"
+          path: "googletest"
+          submodules: "recursive"
```
ThomsonTan, line 27: Seems submodule is not needed?

.github/workflows/glog.yml
```diff
@@ -0,0 +1,102 @@
...
+          sudo apt install -y --no-install-recommends --no-install-suggests \
+            build-essential \
+            cmake \
+            ninja-build \
+            libssl-dev \
+            libcurl4-openssl-dev \
+            libprotobuf-dev \
+            protobuf-compiler \
+            libgmock-dev \
+            libgtest-dev \
```
ThomsonTan, line 60: Is the source check our needed as `libgtest-dev` is installed?

marcalff: Yes, a source checkout for googletest 1.12 is required, because libgtest-dev only installs 1.11.

Installing both 1.11 and 1.12 also tests this line:

```
find_package(GTest 1.12 REQUIRED)
```

to make sure asking for 1.12 discards the default libgtest-dev.

.github/workflows/glog.yml
marcalff: Resolved by upgrading to glog 0.7.0, now required, and fixing deprecation warnings.

.github/workflows/glog.yml
marcalff: Not needed, removed.

ThomsonTan (approved):

PR #424, opened 2024-04-04
[CI] Add CI for instrumentation/boost_log

marcalff:
Fixes #420 

Fixed CMakeLists:

* Require GTest 1.12.0. Building with 1.11 breaks the build due to a gtest bug.
* Require Boost.Log 1.73, as documented in the README.md. Building with 1.66 breaks the build, as found locally.

Relevant links:

* https://github.com/google/googletest/releases/tag/release-1.12.0
  * Fixes the GMock issue, but does not advertise 1.12
* https://github.com/google/googletest/releases/tag/release-1.12.1
  * Fixes the GMock issue, and properly advertises 1.12.1

PR #427, opened 2024-04-09
Fix UT when run with ctest or gtest_filter

chusitoo:
Somewhat related to #424 

- Enable testing with ctest
- Fix failing test suite


marcalff (approved):
Thanks for the fix.

PR #428, opened 2024-04-10
Adding missing root span attributes and a Directive to configure root span name (operation name)

aryanishan1001:
Added two features -
1. $remote_port and User-Agent header value was added as a root span attribute.
2. NginxModuleOperationName directive is introduced, it can be used specify root span name directly for endpoints.


DebajitDas (approved):
LGTM

PR #436, opened 2024-05-02
Nginx version update to 1.26.0 stable and 1.25.5 mainline for otel webserver agent

aryanishan1001:
The Opentelemetry webserver module is now built for Nginx stable version 1.26 and mainline version 1.25.5.


DebajitDas (approved):
LGTM

PR #441, opened 2024-05-10
Upgrade exporters/fluentd to opentelemetry-cpp 1.15.0

marcalff:


marcalff:
Looks like there is no CI in the repo.

Did not test locally, can't even get cmake to build.

marcalff:
Closing, no longer relevant now that opentelemetry-cpp 1.19 is used.

PR #443, opened 2024-05-15
fix(nginx): set error status on span with 5xx status code

dmehala:
## Description
This PR enhances the `otel_ngx_module` to set the span status as "error" specifically when the HTTP request returns a `5xx` status code. For all other status codes, it sets the status as "ok".

### DIFF
--- a/instrumentation/nginx/src/otel_ngx_module.cpp
+++ b/instrumentation/nginx/src/otel_ngx_module.cpp
@@ -545,7 +545,14 @@ ngx_int_t FinishNgxSpan(ngx_http_request_t* req) {
   }
 
   auto span = context->request_span;
-  span->SetAttribute("http.status_code", req->headers_out.status);
+  const auto status_code = req->headers_out.status;
+  span->SetAttribute("http.status_code", status_code);
+
+  if (status_code >= 500) {
+    span->SetStatus(trace::StatusCode::kError);
+  } else {
+    span->SetStatus(trace::StatusCode::kOk);
+  }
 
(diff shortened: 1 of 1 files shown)


dmehala:
Hello,

Do you have an estimated timeline? My understanding is that the `otel-webserver-module` is the new implementation. However, if I am not mistaken, this module is used by [ingress-nginx](https://github.com/kubernetes/ingress-nginx) to provide tracing, and this fix would benefit the entire community.

instrumentation/nginx/src/otel_ngx_module.cpp
```diff
@@ -545,7 +545,14 @@ ngx_int_t FinishNgxSpan(ngx_http_request_t* req) {
   }
 
   auto span = context->request_span;
-  span->SetAttribute("http.status_code", req->headers_out.status);
+  const auto status_code = req->headers_out.status;
+  span->SetAttribute("http.status_code", status_code);
+
+  if (status_code >= 500) {
```
tobiasstadler, line 551: Isn’t 400 also an error?

dmehala: 4xx HTTP status code are client error code. I don't think it is fair to consider the whole span problematic when the issue is from the user.

tobiasstadler: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status says you are right

tobiasstadler (approved):

PR #455, opened 2024-06-10
[user_events] Configure the provider name via exporter option

ThomsonTan:


Pravila00:
For reference, these changes work great within my C++ application. 
I can see new LinuxTracepoint files using the provider name I defined.

lalitb (approved):
LGTM.

PR #456, opened 2024-06-11
Use correct library name in examples

bergjohan:


marcalff (approved):
LGTM, thanks for the fix.

PR #460, opened 2024-07-04
[BOOST LOG] Use correct type for severity logger

### LINKED ISSUE #453: Use correct type for severity_logger
**Steps to reproduce**
Create a severity_logger with the type of `boost::log::trivial::severity_level` instead of `int`.

```cpp
boost::log::sources::severity_logger<boost::log::trivial::severity_level> logger;
BOOST_LOG_SEV(logger, boost::log::trivial::info) << "Test message with severity";
```

**What is the expected behavior?**
```
  severity_num       : 9
  severity_text      : INFO
  body               : Test message with severity
```

**What is the actual behavior?**
```
  severity_num       : 0
  severity_text      : INVALID
  body               : Test message with severity
```

**Additional context**
`boost::log::extract_or_default` fails here, because it expects the type of `int`: https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/boost_log/src/sink.cc#L71

We are clearly using `boost::log::trivial::severity_level` by default (https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/boost_log/include/opentelemetry/instrumentation/boost_log/sink.h#L60), so why not use the correct type everywhere?

chusitoo:
Resolves #453

### DIFF
--- a/instrumentation/boost_log/README.md
+++ b/instrumentation/boost_log/README.md
@@ -33,14 +33,14 @@ boost::log::core::get()->add_sink(sink);
 
 This will create a backend with the following assumptions about the attributes it collects, more precisely:
 
-| Keyword      | Type                        |
-|--------------|-----------------------------|
-| Severity     | int                         |
-| TimeStamp    | boost::posix_time::ptime    |
-| ThreadID     | boost::log::aux::thread::id |
-| FileName     | std::string                 |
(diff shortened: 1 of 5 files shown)


marcalff (approved):
LGTM, thanks for the fix.

PR #497, opened 2024-12-05
[Geneva] Inlcucde UDS header for Linux only

ThomsonTan:


exporters/geneva/src/exporter.cc
```diff
@@ -3,9 +3,10 @@
 
 #include "opentelemetry/exporters/geneva/metrics/exporter.h"
 #include "opentelemetry/exporters/geneva/metrics/macros.h"
-#include "opentelemetry/exporters/geneva/metrics/unix_domain_socket_data_transport.h"
 #ifdef _WIN32
 #include "opentelemetry/exporters/geneva/metrics/etw_data_transport.h"
+#else
+#include "opentelemetry/exporters/geneva/metrics/unix_domain_socket_data_transport.h"
```
ThomsonTan, line 9: In theory, UDS can work on Windows, but I don't think there is any real user scenario. Probably exclude it from Windows build for now.

lalitb (approved):

PR #503, opened 2024-12-06
Support for publishing resource attirbutes in fluent logger.

lalitb:
This PR Incorporate changes from #500, as @bnavarma is experiencing issues with CLA signing. The changes are relevant and are gated behind a feature flag."

### DIFF
--- a/exporters/fluentd/include/opentelemetry/exporters/fluentd/common/fluentd_common.h
+++ b/exporters/fluentd/include/opentelemetry/exporters/fluentd/common/fluentd_common.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "opentelemetry/common/attribute_value.h"
+#include "opentelemetry/sdk/common/attribute_utils.h"
 #include "opentelemetry/exporters/fluentd/common/fluentd_logging.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/version.h"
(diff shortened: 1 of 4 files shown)


marcalff (changes requested):
Please clarify:

> This PR incorporate changes from https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/500, as @bnavarma is experiencing issues with CLA signing.

This would taint the contrib repository if I understand correctly,
because content is getting in, without a CLA.

lalitb:
> This would taint the contrib repository if I understand correctly,
> because content is getting in, without a CLA.

I don’t think that’s the case. The change is covered by my CLA, as I am raising the PR. Just for a background - This exporter is maintained by @ThomsonTan and me, and it’s used internally within Microsoft. @bnavarma is one of the users of this within MS :)

marcalff:
> @bnavarma is one of the users of this within MS :)

Ok then.

exporters/fluentd/include/opentelemetry/exporters/fluentd/common/fluentd_common.h
```diff
@@ -99,6 +100,93 @@ void inline PopulateAttribute(
   }
 }
 
+void inline PopulateOwnedAttribute(
```
ThomsonTan, line 103: nit: wrap this function `PopulateOwnedAttribute` in the new experimental macro, or we'll get an unused function for build without enabling the experimental flag?

ThomsonTan (approved):
Looks good to me with a suggestion.

PR #533, opened 2025-03-11
[user_events] Set EventHeader name to the event_name property

ThomsonTan:
With below PR in main, the user_events Log exporter could create `EventBuilder` during calling into `Recordable` with `event_name`, and also no need to cache the Log attributes which would require extra dynamic memory allocation.

https://github.com/open-telemetry/opentelemetry-cpp/pull/3296

### DIFF
--- a/exporters/user_events/src/recordable.cc
+++ b/exporters/user_events/src/recordable.cc
@@ -104,10 +104,17 @@ void Recordable::SetTimestamp(opentelemetry::common::SystemTimestamp timestamp) 
 
 bool Recordable::PrepareExport() noexcept
 {
-  if (cs_part_b_bookmark_size_ > 0)
+  if (cs_part_b_bookmark_size_ == 0)
+  {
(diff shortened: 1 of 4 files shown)


exporters/user_events/src/recordable.cc
```diff
@@ -17,50 +17,58 @@ namespace logs
...
 {
-  // Set intial bookmark size to 1 for body below.
-  cs_part_b_bookmark_size_++;
+  if (severity_ == 0)
+  {
+    OTEL_INTERNAL_LOG_ERROR("[user_events Log Exporter] Recordable: severity is not set.");
+    return;
+  }
+
+  auto event_name = !event_name_.empty() ? event_name_.data() : "OpenTelemetryLogs";
```
lalitb, line 44: Rust uses default as "Logs". Either is fine, as long as we are consistent. cc @cijothomas.

lalitb (approved):
LGTM. With nit comment. Thanks for the PR.

PR #547, opened 2025-04-30
[user_events] Write _typeName to event builder directly

ThomsonTan:
Avoid calling `utils::PopulateAttribute()` for "_typeName" because its type is well-known.


lalitb (approved):
with changes requested.

PR #540, opened 2025-03-27
Add SocketDataTransport to support kTCP/kUDP

zurex:
## Description
This PR raised for fix #536 , navigate to the issue to get more details.

## Changes
* Add new `SocketDataTransport` to support UNIX/UDP/TCP protocol.
* Remove existing `UnixDomainDataTransport` since `SocketDataTransport` covered unix scenario.
* Update the version of `geneva-exporter` to `1.0.1`.

## Test && Validation
* Test in production scenarios.
* Passed the test cases.

## Existing Behavior
`exporter` only support `kETW` and `kUNIX` protocol although the connection string support `kUDP` and `kTCP`.

## New Behavior
Support all the protocols which defined in [TransportProtocol](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/connection_string_parser.h#L22):
* kETW
* kUNIX
* kUDP (new added)
* kTCP (new added)

PR #549, opened 2025-05-09
Revert "Add SocketDataTransport to support kTCP/kUDP (#540)"

ThomsonTan:
This reverts commit 2184b8341073930d0b6926b6411147bb3ad6992c.

There is no plan to support kTCP/kUDP in the near future. Will log error when these protocols are used.

FYI @zurex

### DIFF
--- a/exporters/geneva/CMakeLists.txt
+++ b/exporters/geneva/CMakeLists.txt
@@ -22,7 +22,7 @@ endif()
 
 include_directories(include)
 
-set(OTEL_GENEVA_EXPORTER_VERSION 1.0.1)
+set(OTEL_GENEVA_EXPORTER_VERSION 1.0.0)
 set(OTEL_GENEVA_EXPORTER_MAJOR_VERSION 1)
 
 function(set_target_version target_name)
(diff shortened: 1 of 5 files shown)


lalitb (approved):

malkia:
Can anyone explain what happened here? This was added first, then removed, but no further notes about why?

lalitb:
> Can anyone explain what happened here? This was added first, then removed, but no further notes about why?

The exporter in this directory is designed only for sending metrics to a locally running collector agent - so it supports UDS on Linux and ETW on Windows. For sending metrics to remote collector, it’s best to use the standard OTLP exporter.

ThomsonTan:
> Can anyone explain what happened here? This was added first, then removed, but no further notes about why?

Yes, as Lalit mentioned, we recommend running this exporter side-by-side with the application. To reduce the confusion, and as mentioned in the revert description, we don't plan to support the remote protocol like TCP or UDP.

I will remove the protocol definition of `kTCP` and 'kUDP` to make this more clear.

PR #553, opened 2025-06-10
[Geneva] Make EXPORT_NAME consistent

ThomsonTan:
Address the commend on inconsistent EXPORT_NAME for different build types of the contrib exportes.

### DIFF
--- a/exporters/fluentd/CMakeLists.txt
+++ b/exporters/fluentd/CMakeLists.txt
@@ -117,6 +117,14 @@ if (WITH_EXAMPLES)
 endif()
 
 if(OPENTELEMETRY_INSTALL)
+  set_target_properties(
+    opentelemetry_exporter_geneva_trace
+    PROPERTIES
+      EXPORT_NAME opentelemetry_exporter_geneva_trace)
+  set_target_properties(
+    opentelemetry_exporter_geneva_logs
+    PROPERTIES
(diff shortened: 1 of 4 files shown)


exporters/fluentd/CMakeLists.txt
```diff
@@ -117,6 +117,14 @@ if (WITH_EXAMPLES)
 endif()
 
 if(OPENTELEMETRY_INSTALL)
```
dbarker, line 119: The export name properties are good to set unconditionally. These lines can go immediately after add_library calls for each target.

exporters/geneva/CMakeLists.txt
```diff
@@ -91,40 +91,36 @@ if(BUILD_TESTING)
...
-
-else()
-
-  set_target_properties(opentelemetry_exporter_geneva_metrics PROPERTIES EXPORT_NAME opentelemetry_exporter_geneva_metrics)
-
-  otel_add_component(
-    COMPONENT
-    exporters_geneva_metrics
-    TARGETS
+  set_target_properties(
```
dbarker, line 94: can set these properties unconditionally immediately after add_library.

exporters/fluentd/CMakeLists.txt
ThomsonTan: Moved it out to be after `add_libary()`.

lalitb (approved):

Issue #232, opened 2022-11-03
nginx: Configure custom resource attributes


DrLuke:
As a user I'd like to define custom resource attributes (like [`deployment.environment`](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/nginx/test/conf/otel-nginx.toml)) with both the [module directives](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/bd4a9c169414e5b1140920ee283fd9e0f0e079f5/instrumentation/otel-webserver-module/src/nginx/ngx_http_opentelemetry_module.c#L167) as well as the [config toml file](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/nginx/test/conf/otel-nginx.toml).

NicklasWallgren:
Has there been any progress around this, or workarounds?

PR #566, opened 2025-07-28
Add opentelemetry_resource_attr

yuriolisa:
- Add opentelemetry_resource_attr

- Resolves: #232


yuriolisa:
@open-telemetry/cpp-contrib-approvers, would it be possible to re-run only the failed jobs?

PR #606, opened 2026-03-16
Fix ODR violation when using Geneva logs and metrics exporters together on Linux

legenapk-ms:
Problem
When including both Geneva exporter headers for logs and metrics in the same translation unit on Linux, the build fails with an ODR (One Definition Rule) violation:
#include "opentelemetry/exporters/geneva/geneva_logger_exporter.h"
#include "opentelemetry/exporters/geneva/metrics/exporter.h"

error: redefinition of 'struct net::Thread'
note: previous definition of 'struct net::Thread'Root Cause: 
Both fluentd/common/socket_tools.h (used by Geneva logger) and geneva/metrics/socket_tools.h (used by Geneva metrics) define identical types (net::Thread, SocketTools::Socket, etc.) in the same global namespaces, causing ODR violations when both headers are included

Solution
Wrapped all socket tools definitions in geneva/metrics/socket_tools.h in a unique geneva_metrics namespace to avoid conflicts:

net::Thread → geneva_metrics::net::Thread
SocketTools::Socket → geneva_metrics::SocketTools::Socket
Added using namespace geneva_metrics; in dependent headers for transparent usage

### DIFF
--- a/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
+++ b/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
@@ -111,8 +111,12 @@ typedef u_short sa_family_t;
 
 #include "opentelemetry/exporters/geneva/metrics/macros.h"
 
-// Wrap in geneva_metrics namespace to avoid ODR violations with fluentd socket_tools.h
-namespace geneva_metrics {
+// Wrap in detail namespace to avoid ODR violations with fluentd socket_tools.h
+namespace opentelemetry {
+namespace exporter {
+namespace geneva {
+namespace metrics {
+namespace detail {
 
 namespace net {
 
@@ -1133,7 +1137,11 @@ public:
 
 } // namespace SocketTools
 
-} // namespace geneva_metrics
+} // namespace detail
+} // namespace metrics
+} // namespace geneva
+} // namespace exporter
+} // namespace opentelemetry
 
 #ifdef _MSC_VER
 #pragma warning(pop)
--- a/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
+++ b/exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
@@ -110,9 +110,10 @@ typedef u_short sa_family_t;
 #endif
 
 #include "opentelemetry/exporters/geneva/metrics/macros.h"
+#include "opentelemetry/version.h"
 
 // Wrap in detail namespace to avoid ODR violations with fluentd socket_tools.h
-namespace opentelemetry {
+OPENTELEMETRY_BEGIN_NAMESPACE
 namespace exporter {
 namespace geneva {
 namespace metrics {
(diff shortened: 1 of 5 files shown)


exporters/geneva/include/opentelemetry/exporters/geneva/metrics/unix_domain_socket_data_transport.h
```diff
@@ -14,6 +14,10 @@ OPENTELEMETRY_BEGIN_NAMESPACE
 namespace exporter {
 namespace geneva {
 namespace metrics {
+
+// Use geneva_metrics namespace for socket tools to avoid conflicts with fluentd
+using namespace geneva_metrics;
```
ThomsonTan: Avoid using namespace in namespace because it expose all the geneva_metrics::* as unqualified names to all the users of this namespace.

exporters/geneva/include/opentelemetry/exporters/geneva/metrics/socket_tools.h
```diff
@@ -111,6 +111,9 @@ typedef u_short sa_family_t;
 
 #include "opentelemetry/exporters/geneva/metrics/macros.h"
 
+// Wrap in geneva_metrics namespace to avoid ODR violations with fluentd socket_tools.h
+namespace geneva_metrics {
```
lalitb: Rather than introducing a new top-level `geneva_metrics` namespace, we can nest this under the existing project
namespace hierarchy to stay idiomatic with the rest of the codebase. For example:

```cpp
 namespace opentelemetry {
 namespace exporter {
 namespace geneva {
 namespace metrics {
 namespace detail {
 
 namespace net { ... }
 namespace SocketTools { ... }
 
 } // namespace detail
 } // namespace metrics
 } // namespace geneva
 } // namespace exporter
 } // namespace opentelemetry
```

exporters/geneva/test/common/socket_server.h
```diff
@@ -38,14 +38,14 @@
 #include "opentelemetry/exporters/geneva/metrics/macros.h"
 #include "opentelemetry/exporters/geneva/metrics/socket_tools.h"
 
-using namespace SocketTools;
-
 namespace SOCKET_SERVER_NS {
 
+namespace ST = geneva_metrics::SocketTools;
```
lalitb: short alias like `ST` is not very descriptive. Something like `socket_tools` or `socket_util` would be clearer at a glance.

exporters/geneva/test/metrics_exporter_test.cc
```diff
@@ -314,8 +314,8 @@ TEST_P(GenericMetricsExporterTextFixture, BasicTests)
   bool isRunning              = true;
 
   // Start test server
-  SocketAddr destination(kUnixDomainPath.data(), true);
-  SocketParams params{AF_UNIX, SOCK_STREAM, 0};
+  geneva_metrics::SocketTools::SocketAddr destination(kUnixDomainPath.data(), true);
```
ThomsonTan: The test still uses `geneva_metrics::SocketTools`, but that intermediate namespace no longer exists after the final refactor. The new  namespace is `opentelemetry::v1::exporter::geneva::metrics::detail::SocketTools`

ThomsonTan (approved):

lalitb (approved):

PR #615, opened 2026-05-11
Add HTTP/protobuf OTLP export support for Apache and Nginx modules

### LINKED ISSUE #614: Add HTTP/protobuf OTLP export support to otel-webserver-module
## Summary

Add support for HTTP/protobuf OTLP export protocol alongside the existing gRPC export in the otel-webserver-module (used for Apache HTTPD and Nginx auto-instrumentation).

## Motivation

The OpenTelemetry Operator provides auto-instrumentation for multiple languages (Java, Python, Node.js, .NET, Go). All other language auto-instrumentations support configuring the OTLP export protocol via the `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable, and default to `http/protobuf`. The Apache HTTPD and Nginx auto-instrumentations are the exception — they hardcode gRPC at the native C++ level by directly linking `libopentelemetry_exporter_otlp_grpc.so`.

This creates operational friction:
- Users must ensure their collector exposes the gRPC endpoint (port 4317) specifically for these two instrumentations, even if everything else in their environment uses HTTP/protobuf (port 4318)
- The `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable is silently ignored, leading to confusion
- Operator users cannot use a uniform collector configuration across all auto-instrumented workloads

## Proposed Change

1. **Link both OTLP exporter libraries** — Build the webserver module against both `opentelemetry_exporter_otlp_grpc` and `opentelemetry_exporter_otlp_http`
2. **Add runtime protocol selection** — Read `OTEL_EXPORTER_OTLP_PROTOCOL` (and the signal-specific variants `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, etc.) at exporter initialization time
3. **Use the appropriate factory** — Instantiate `OtlpHttpExporterFactory` or `OtlpGrpcExporterFactory` based on the configured protocol
4. **Default to `http/protobuf`** — Align with the OpenTelemetry specification default and the behavior of all other auto-instrumentations

This is architecturally straightforward because the upstream opentelemetry-cpp SDK already provides both exporters behind a shared `SpanExporter` interface. The webserver module pipeline is transport-agnostic — only the exporter instantiation point needs to change.

## Scope

- Apache HTTPD auto-instrumentation
- Nginx auto-instrumentation
- CMake/build changes to link the HTTP exporter library and its dependency (libcurl)
- Supported protocol values: `grpc`, `http/protobuf`, `http/json`

## Related

- Discussion on separating OTLP proto from transport: https://github.com/open-telemetry/opentelemetry-cpp-contrib/discussions/410
- OpenTelemetry specification on `OTEL_EXPORTER_OTLP_PROTOCOL`: https://opentelemetry.io/docs/specs/otel/protocol/exporter/

pavolloffay:
This change adds runtime OTLP protocol selection for the otel-webserver-module,                                                                                                                                                                                                                        
  allowing users to choose between grpc, http/protobuf, and http/json exporters.
                                                                                                                                                                                                                                                                                                         
  - Link both libopentelemetry_exporter_otlp_grpc.so and                                                                                                                                                                                                                                                 
    libopentelemetry_exporter_otlp_http.so                                                                                                                                                                                                                                                               
  - Add ApacheModuleOtelExporterOtlpProtocol and                                                                                                                                                                                                                                                         
    NginxModuleOtelExporterOtlpProtocol configuration directives                                                                                                                                                                                                                                         
  - Support OTEL_EXPORTER_OTLP_TRACES_PROTOCOL and OTEL_EXPORTER_OTLP_PROTOCOL
    environment variable fallbacks                                                                                                                                                                                                                                                                       
  - Default to http/protobuf per OTel specification         
                                                                                                                                                                                                                                                                                                         
  Fixes #614


pavolloffay:
@chrlic could you please review?

chrlic:
It looks good to me, very useful addition.

lalitb:
Thanks @pavolloffay for the PR - 

Before merging this, can we fix/refresh the webserver CI and dependency versions? This PR changes code that should be compiled against the webserver module’s pinned dependencies, but the current CI does not appear reliable because some of the runner/distro setup is outdated. At the same time, the module still advertises/builds against older pinned versions, including [opentelemetry-cpp 1.2.0](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/otel-webserver-module/version.properties).

It would be safer to first update the CI to a supported runner/distro setup and make sure the module builds against the intended OpenTelemetry C++ SDK version, then validate this PR against that. Otherwise we may merge API usage or runtime behavior that is not actually compatible with the versions this module packages.

Also as mentioned in slack - the modules is currently not-maintained, and we are looking for the maintainers for this.

lalitb:
@pavolloffay  - Just checking if you are still working on this PR?

instrumentation/otel-webserver-module/build.gradle
```diff
@@ -189,6 +189,8 @@ task stageLibrary(type: Copy) {
     from("${modDepDir}/opentelemetry/${cppSDKVersion}/lib/libopentelemetry_otlp_recordable.so") { it.into "sdk_lib/lib" }
     from("${modDepDir}/opentelemetry/${cppSDKVersion}/lib/libopentelemetry_exporter_ostream_span.so") { it.into "sdk_lib/lib" }
     from("${modDepDir}/opentelemetry/${cppSDKVersion}/lib/libopentelemetry_exporter_otlp_grpc.so") { it.into "sdk_lib/lib" }
+    from("${modDepDir}/opentelemetry/${cppSDKVersion}/lib/libopentelemetry_exporter_otlp_http.so") { it.into "sdk_lib/lib" }
+    from("${modDepDir}/opentelemetry/${cppSDKVersion}/lib/libopentelemetry_http_client_curl.so") { it.into "sdk_lib/lib" }
```
proost, line 193: We need to SDK version to 1.3.0. See this: https://github.com/open-telemetry/opentelemetry-cpp/releases/tag/v1.3.0

proost (commented):
We need double check that current change is compatible with otel-cpp 1.2.0

PR #619, opened 2026-06-22
[geneva/metrics] prevent buffer overflow in attribute serialization

ThomsonTan:
`SerializeString`/`SerializeInt` wrote into the fixed `buffer_[kBufferSize]` (65360 bytes) with no bounds check, so attributes whose total serialized size exceeded the buffer overflowed it). The declared `kMaxDimensionValueSize` (1024) cap was never enforced on values, and the `uint16_t` length prefix was stored via an unaligned `reinterpret_cast`.

### DIFF
--- a/exporters/geneva/src/exporter.cc
+++ b/exporters/geneva/src/exporter.cc
@@ -105,8 +105,10 @@ opentelemetry::sdk::common::ExportResult Exporter::Export(
               sdk::metrics::AggregationType::kSum, event_type, new_value,
               metric_data.end_ts, metric_data.instrument_descriptor.name_,
               point_data_with_attributes.attributes);
-          data_transport_->Send(event_type, buffer_,
-                                body_length + kBinaryHeaderSize);
+          if (body_length > 0) {
+            data_transport_->Send(event_type, buffer_,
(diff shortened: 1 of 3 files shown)


exporters/geneva/src/exporter.cc
lalitb: do we need space check for these SerializeInt too ?

lalitb (approved):
LGTM. Thanks for fixing this. Probably, we also need to tighten the histogram path - we currently ignore the SerializeInt return value there. Also  good to add a histogram regression test with enough non-empty buckets to hit this path.

exporters/geneva/src/exporter.cc
ThomsonTan: The space check is added for `SerializeInt` in the PR, as below. Or let me know if anything is missed.

```c++
static bool SerializeInt(char *buffer, size_t &index, T value) {
  if (index + sizeof(T) > kBufferSize) {
    return false;
  }
```

lalitb: Yes, the check is inside SerializeInt. My concern is that this call site ignores the false return value.

ThomsonTan: Added check on the return value in necessary places.

PR #628, opened 2026-07-20
[spdlog] fix type conversion error

rafe-murray:
Hi there, I was planning on using the spdlog sink, but I noticed it wasn't compiling on my machine. It seems like on some systems `opentelemetry::common::AttributeValue` cannot accept `size_t`. 

Looking at what clang is deducing the types as it seems the issue is that uint32_t evaluates to `unsigned int`, uint64_t to `unsigned long long`, but size_t to `unsigned long`. 

Since `size_t` is only _guaranteed_ to hold at least 32 bits, I'm proposing we use a static cast to `uint32_t`. Alternatives would be casting to `uint64_t` (which is not officially supported in the standard but would never lose information) or splitting it into a `nostd::span` if the platform has a 64-bit size_t implementation.

For context I am compiling on an M2 mac with macOs 15.7.4 using LLVM clang.

### DIFF
--- a/instrumentation/spdlog/src/sink.cc
+++ b/instrumentation/spdlog/src/sink.cc
@@ -37,7 +37,7 @@ void OpenTelemetrySink<Mutex>::sink_it_(const spdlog::details::log_msg &msg)
       log_record->SetAttribute(kCodeFilePath, msg.source.filename);
       log_record->SetAttribute(kCodeLineNumber, msg.source.line);
     }
-    log_record->SetAttribute(kThreadId, msg.thread_id);
+    log_record->SetAttribute(kThreadId, static_cast<uint32_t>(msg.thread_id));
     logger->EmitLogRecord(std::move(log_record));
   }
 }
--- a/instrumentation/spdlog/src/sink.cc
+++ b/instrumentation/spdlog/src/sink.cc
@@ -37,7 +37,7 @@ void OpenTelemetrySink<Mutex>::sink_it_(const spdlog::details::log_msg &msg)
       log_record->SetAttribute(kCodeFilePath, msg.source.filename);
       log_record->SetAttribute(kCodeLineNumber, msg.source.line);
     }
-    log_record->SetAttribute(kThreadId, static_cast<uint32_t>(msg.thread_id));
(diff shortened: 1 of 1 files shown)


instrumentation/spdlog/src/sink.cc
```diff
@@ -37,7 +37,7 @@ void OpenTelemetrySink<Mutex>::sink_it_(const spdlog::details::log_msg &msg)
       log_record->SetAttribute(kCodeFilePath, msg.source.filename);
       log_record->SetAttribute(kCodeLineNumber, msg.source.line);
     }
-    log_record->SetAttribute(kThreadId, msg.thread_id);
+    log_record->SetAttribute(kThreadId, static_cast<uint32_t>(msg.thread_id));
```
marcalff:
```suggestion
    log_record->SetAttribute(kThreadId, static_cast<uint64_t>(msg.thread_id));
```

marcalff:
@rafe-murray 

Thanks for the fix.

> Alternatives would be casting to `uint64_t` (which is not officially supported in the standard but would never lose information)

I changed the fix to use `uint64_t` instead, it was expected by existing unit tests:

```
Note: Google Test filter = OpenTelemetrySinkTest.Log_Success
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from OpenTelemetrySinkTest
[ RUN      ] OpenTelemetrySinkTest.Log_Success
/home/runner/work/opentelemetry-cpp-contrib/opentelemetry-cpp-contrib/opentelemetry-cpp-contrib/instrumentation/spdlog/test/sink_test.cc:158: Failure
Value of: nostd::holds_alternative<uint64_t>(thread_id)
  Actual: false
Expected: true
[  FAILED  ] OpenTelemetrySinkTest.Log_Success (0 ms)
[----------] 1 test from OpenTelemetrySinkTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] OpenTelemetrySinkTest.Log_Success

 1 FAILED TEST
```

marcalff (approved):
LGTM, thanks for the fix.

PR #633, opened 2026-07-25
ci: Resolving nginx & prometheus build issue + web server consistency

thompson-tomo:
NGINX instrumentation build failures have been addressed by explicitly using buildx & switching tests to use php docker image which exists. Have also updated the tests to use nil for remote-parent-id rather than empty string.

Prometheus bazel build resolved by specifying an older bazel version to use.

Prometheus cmake build resolved by changed zlib adding process.

Httpd/webserver resolved by updating runner image.

This restricts all push triggers to main to improve ci consistency and remove unnecessary checks on pr’s. To build for a commit use the manual trigger or use a draft pr.


.github/workflows/webserver.yml
```diff
@@ -196,31 +196,3 @@ jobs:
       #     sleep 30
       #     ./gradlew :test:integration:integrationTests -i
       #     curl http://localhost:9411/api/v2/spans?serviceName=demoservice
-
-
-  Codeql-build:
```
lalitb, line 201: This is removed - should be fine if we want to bring it back as followup. Just want to ensure it's intentional.

lalitb:
Thanks for improving the CI here - most of the previously failing jobs are now green. One Prometheus issue remains: the Linux CMake job still fails while configuring the upstream `opentelemetry-cpp` checkout because `prometheus-cpp::pull` references a missing `ZLIB::ZLIB` target. The new `find_package(ZLIB REQUIRED)` is in the contrib CMakeLists, which is configured later, so it cannot affect this failure. Could you please address this remaining case and get the job green?


```
-- Using external prometheus-cpp
-- Building WITH_STL=OFF
-- Configuring done (1.7s)
CMake Error at /usr/local/share/vcpkg/installed/x64-linux/share/prometheus-cpp/prometheus-cpp-targets.cmake:76 (set_target_properties):
  The link interface of target "prometheus-cpp::pull" contains:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/local/share/vcpkg/installed/x64-linux/share/prometheus-cpp/prometheus-cpp-config.cmake:48 (include)
  /usr/local/share/vcpkg/scripts/buildsystems/vcpkg.cmake:939 (_find_package)
  CMakeLists.txt:369 (find_package)


CMake Warning:
-- Generating done (0.0s)
  Manually-specified variables were not used by the project:

    WITH_METRICS_PREVIEW


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

.github/workflows/prometheus.yml
```diff
@@ -18,7 +19,7 @@ permissions:
 jobs:
   prometheus_bazel_linux:
     name: Bazel on Linux
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
owent: Why not use the latest build environment?

thompson-tomo: This way it is explicitly controlled the version we are on & we will get a renovate pr to bump it when a new one is available.

.github/workflows/webserver.yml
thompson-tomo: Yes it is intentional as it appears to have never worked & is being re-added via separate pr which ensures it works.

thompson-tomo:
@lalitb i am out of ideas as to what to do to get it green given my knowledge of these types of builds.

lalitb:
> @lalitb i am out of ideas as to what to do to get it green given my knowledge of these types of builds.

@thompson-tomo - The Prometheus workflow  performs two separate CMake configurations:
1. It first configures upstream `opentelemetry-cpp`.
2. It then configures this contrib Prometheus exporter.

The failure happens during the first configuration, when `prometheus-cpp` tries to use `ZLIB::ZLIB`.

 This PR adds `find_package(ZLIB REQUIRED)` to the contrib CMakeLists, but that file is used only during the second configuration. Therefore, the command runs too late to fix the first configuration.

Could you make ZLIB available before the first CMake command that enables Prometheus?

proost:
@thompson-tomo 
For the Webserver centos7, we need to update command in the dockerfile. see [this PR](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/645/changes#diff-83a6e210a3ec60f07db2787f9ae3d3fb552247db72395209881bab1b7f2c38f9)

thompson-tomo:
@proost i would suggest you cherrypick the changes to https://github.com/thompson-tomo/opentelemetry-cpp-contrib/blob/ci_fix/.github/workflows/webserver.yml onto your branch which will enable the ci to run in your branch. I will then remove them from here to avoid conflicts.

thompson-tomo:
@lalitb / @dbarker any further guidance on how to resolve https://github.com/open-telemetry/opentelemetry-cpp-contrib/actions/runs/32260993486/job/96093945611#step:5:1

thompson-tomo:
Please see #667 for continuation of this pr

PR #634, opened 2026-07-25
ci: Add dedicated CodeQL workflow for improved analysis

thompson-tomo:
This adds a dedicated codeql workflow to resolve the issue with the current ci failures where in which codeql is tied to a non c++ project.

Nginix & httpd is excluded and will be enabled via #633 & #638


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,64 @@
...
+            ninja-build \
+            libssl-dev \
+            libcurl4-openssl-dev \
+            libprotobuf-dev \
+            protobuf-compiler \
+            libgmock-dev \
+            libgtest-dev
+      - name: Build and upload package
+        run: |
+          for conan_file in $(find exporters instrumentation -name conanfile.txt); do
```
dbarker: Conan may be a bit too heavy and the otel-cpp release availability lag too much for the CI workflows.

Alternatively we can directly clone and build the `opentelemetry-cpp` dependency.  This gives us more control over what we test and keep it up to date with the otel-cpp releases (less of a concern for codeql, but will be important for other ci workflows).  

Instrumentation libraries should only need the API, while exporters and examples only need the API + SDK so the default options should do fine.

```sh
git clone --recurse-submodules --depth 1 --branch v1.28.0 \
  https://github.com/open-telemetry/opentelemetry-cpp.git
cd opentelemetry-cpp
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DOPENTELEMETRY_INSTALL=ON
cmake --build build --target install --parallel
```

instrumentation/glog/conanfile.txt
```diff
@@ -0,0 +1,7 @@
+[requires]
```
dbarker: Adding conanfiles to each contrib package seems to imply support for conan builds (and maintenance of them), which isn't strictly needed by codeql (a suitable apt package will work). Package manager support (conan, vcpkg, etc) are community driven for the main repo and we should keep package manager agnostic in contrib as well.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          sudo apt update -y
+          sudo apt install -y --no-install-recommends --no-install-suggests \
+            build-essential \
+            cmake \
+            ninja-build \
+            libssl-dev \
+            libcurl4-openssl-dev \
+            libprotobuf-dev \
+            protobuf-compiler \
```
dbarker: Please try with the apt packages for the dependencies in place of conan. See related comment below.  

nlohmann-json3-dev 
libgoogle-glog-dev
libboost-log-dev
libspdlog-dev

dbarker (commented):
Thanks for contributing to CI. It is needed! Just a request to use native CMake install for otel-cpp and apt packages for the dependencies so conan support is not implied and it is easier maintain.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,64 @@
...
+  contents: read
+
+jobs:
+  Codeql-build:
+    permissions:
+      security-events: write # for github/codeql-action/analyze to upload SARIF results
+    name: static-analysis
+    runs-on: ubuntu-24.04
+    steps:
+      - name: checkout otel contrib
```
dbarker, line 58: please add the harden-runner step and pin all steps to the correct SHA (copied below from the otel-cpp repo's codeql workflow). 

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/.github/workflows/codeql-analysis.yml

```
 - name: Harden the runner (Audit all outbound calls)
      uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
      with:
        egress-policy: audit

  - name: Checkout repository
    uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
    with:
       submodules: 'recursive'

 - name: Initialize CodeQL
   uses: github/codeql-action/init@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3

 - name: Perform CodeQL Analysis
   uses: github/codeql-action/analyze@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3
```

dbarker (commented):
Please see below on steps to harden the workflow..

.github/workflows/codeql.yml
thompson-tomo: The problem with using apt packages is that there is no way to specify that this package needs dependency xyz and is a nightmare if we wanted control versions for reproducible builds.

Conanfiles solve all that especially as we can use renovate to propose pr's to update them.

You could possibly solve it by custom tooling but conanfiles is an easier to maintain solution.

.github/workflows/codeql.yml
thompson-tomo: The clone option didn't work as different packages needed different versions. In relation to release lag, i am not sure what you mean. If it is the time to update the version in the file, that is solved by having renovate update the dependency like almost all other otel repos.

> Instrumentation libraries should only need the API, while exporters and examples only need the API + SDK so the default options should do fine.

As per spec these should be seperate packages https://opentelemetry.io/docs/specs/otel/library-guidelines/#expected-usage

instrumentation/glog/conanfile.txt
thompson-tomo: See https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/634#discussion_r3670381593 but apt-packages introduce more problems. The building of packages is out of scope it is purely a dependency management solution.

.github/workflows/codeql.yml
dbarker: > different packages needed different versions. 
 
This needs to be addressed first. The opentelemetry-cpp API is stable with ABIv1 and there should not be a version compatibility issue with instrumentation components. Maintaining different otel-cpp versions for each contrib component will not be sustainable and is an unnecessary burden on maintainers. 

For those contrib components that rely on the SDK (exporters, examples) we should first try to see if they all build with the latest 1.28 release. 

For this repo to be sustainable going forward we need code owners for the components who can maintain them. Building and passing tests against the latest otel-cpp version seems like a reasonable minimum bar to start with.  

Please keep in mind that this repo hasn't received a lot of attention over the years and some components likely can/should be removed from the build and CI as no longer maintained. There is no policy written down for this at the moment but we may take inspiration from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#becoming-a-code-owner. 

> In relation to release lag, i am not sure what you mean.
 
The latest conan `opentelemetry-cpp` package is for version 1.26 (https://conan.io/center/recipes/opentelemetry-cpp), while the current release is 1.28. The conan package is supported by the community (which is awesome!) and is updated at its own pace. 

For CI testing of `opentelemetry-cpp-contrib` we need to be able to test against the latest opentelemetry-cpp release (and potentially against the latest HEAD of `main` on some regular cadence) to detect regressions. This would support a path towards releasing versions of otel-cpp-contrib along side the core otel-cpp repo.  

> As per spec these should be seperate packages

Yes!  opentelemetry-cpp provides the CMake tools to generate API only and API+SDK packages, but does not distribute or release pre-built packages at this time. The packages available through common package managers are all community supported and best effort. Please see:  https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#generating-binary-packages

.github/workflows/codeql.yml
dbarker: Makes sense. In opentelemetry-cpp's codeql workflow apt packages are being used since their versions all fall within the minimum supported range. Explicitly installing pinned versions is better. Ideally the approach  taken is consistent for all dependencies installed within the workflow.  

An alternative to conan is to use native CMake (ExternalProject, FetchContent). This is the approach in  `opentelemetry-cpp` and the ci scripts can be used in this repo if the dependencies are already supported (an example is [user_events.yml#L54](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/.github/workflows/user_events.yml#L54)).   Please see: 
- https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ci/install_thirdparty.sh
- https://github.com/open-telemetry/opentelemetry-cpp/blob/main/install/cmake/CMakeLists.txt  

The [third_party_release](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release) file defines the current dependency release tags and is updated manually per release. We could use renovate and a contribution for that would be welcome! The format of this file can be made renovate friendly if needed. 

The rationale for native CMake is to stay package manager agnostic and ensure CI is testing against the dependencies' CMake installed package  (or building them as subprojects with FetchContent) as released. Conan changes how a CMake project is configured and installed (patches are applied, target types are changed, new targets are added, etc.). Passing a build/test with a conan installed package is not the same as building against the dependencies installed through their native CMake configuration.

.github/workflows/codeql.yml
thompson-tomo: > The opentelemetry-cpp API is stable with ABIv1 and there should not be a version compatibility issue with instrumentation components. Maintaining different otel-cpp versions for each contrib component will not be sustainable and is an unnecessary burden on maintainers.

I agree i first wanted to just get everything building so we can see what is out of date & fails when updated by renovate.

> For this repo to be sustainable going forward we need code owners for the components who can maintain them. Building and passing tests against the latest otel-cpp version seems like a reasonable minimum bar to start with.

Also agree but i would add that the expectation is that each instrumentation can be built using a standardised ci as otherwise it also unmaintainable & wont scale.

> The conan package is supported by the community (which is awesome!) and is updated at its own pace.

> Yes! opentelemetry-cpp provides the CMake tools to generate API only and API+SDK packages, but does not distribute or release pre-built packages at this time. The packages available through common package managers are all community supported and best effort.

Interesting, i wasn't aware that it wasn't released by the org. It does however appear to violate https://opentelemetry.io/docs/specs/otel/overview/#instrumentation-libraries as users like myself might be mistaken to believing it is an officially released package. Given how important especially the api package should that be released by the org?

> For CI testing of opentelemetry-cpp-contrib we need to be able to test against the latest opentelemetry-cpp release (and potentially against the latest HEAD of main on some regular cadence) to detect regressions. This would support a path towards releasing versions of otel-cpp-contrib along side the core otel-cpp repo.

That can certainly be looked at as a later step. The steps i see:

- Getting it to build in a common manner regardless of os
- Getting tests to run & pass in a common manner regardless of os
- testing with latest sdk to pre-flag issuues

.github/workflows/codeql.yml
thompson-tomo: The first thing which stands out, that script solution only works for linux and is something needing to be maintained, whereas using a package manager is cross platform solution available out of the box.

> The rationale for native CMake is to stay package manager agnostic and ensure CI is testing against the dependencies' CMake installed package (or building them as subprojects with FetchContent) as released. Conan changes how a CMake project is configured and installed (patches are applied, target types are changed, new targets are added, etc.). Passing a build/test with a conan installed package is not the same as building against the dependencies installed through their native CMake configuration.

Understood, is the solution not to do a hybrid of both? In fact it ties into your suggestion above. For instance we use conan for reproducible builds to replicate a real world consumption. This targets latest release. We at the same time use native cmake to compile using latest main and run tests (subset). That way we have best of both worlds.

.github/workflows/codeql.yml
thompson-tomo: Done

.github/workflows/codeql.yml
dbarker: > Interesting, i wasn't aware that it wasn't released by the org. It does however appear to violate https://opentelemetry.io/docs/specs/otel/overview/#instrumentation-libraries as users like myself might be mistaken to believing it is an officially released package. Given how important especially the api package should that be released by the org?

There are many package managers and we cannot maintain recipes/ports for all of them :). So we focus on the source releases now. 

The package managers seem to best fit into the OpenTelemetry Distributions concept: https://opentelemetry.io/docs/concepts/distributions/. They provide recipes/ports that change build options, patch sources, and apply ecosystem-specific toolchains/configuration files on build/install.

> That can certainly be looked at as a later step. The steps i see:
> 
> - Getting it to build in a common manner regardless of os
> - Getting tests to run & pass in a common manner regardless of os
> - testing with latest sdk to pre-flag issuues

I agree. Conan won't fit the role of the common manager for the core workflows though. With Conan, the specific concern is that the recipes often rewrite the config files impacting targets, components, and variables made available through `find_package` compared with native CMake usage. This is the case for the `opentelemetry-cpp` recipe. 

For the `opentelemetry-cpp` dependency and those that are built with CMake we need to use native CMake installs with the `ExternalProject` module in core workflows. 

For the CodeQL workflow we can either use apt packages or native CMake installs if available.

.github/workflows/codeql.yml
dbarker: The Linux script is just a helper that calls CMake. Using CMake’s [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) is our maintainable, cross-platform way to build/install pinned dependencies without adopting a single package manager. The [install/cmake/CMakeLists.txt](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/install/cmake/CMakeLists.txt) CMake project is what we use to install dependencies in the dev container locally (see [opentelemetry-cpp/ci/README.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ci/README.md)) and is used by the CI workflows. 

I agree we should consider a separate/additional Conan-specific workflow later. It just isn't the right fit for core workflows and adding conanfiles to each component sets the wrong precedent before we have CI infrastructure built for this project. Conan should not be required to run CodeQl.

We have some real-world coverage tests for building opentelemetry-cpp with package manager installed dependencies via a dedicated [install workflow](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/.github/workflows/cmake_install.yml) that covers Conan, vcpkg, and Homebrew. These are not the default in core CI workflows and tests for package manager recipes/ports of `opentelemetry-cpp` is not something we can or should pursue.

dbarker (changes requested):
Thanks for taking this on. Following the discussion above please remove the conanfiles and install the dependencies for the CodeQL test with either apt packages or using  CMake's `ExternalProject` module.

lalitb:
Thanks for adding the shared CodeQL workflow. Two concerns:

- I agree with @dbarker that CodeQL should not require adding Conan support to each component. Native CMake or system packages are a better fit here.
-  The workflow triggers for all `exporters/` and `instrumentation/` changes, but only analyzes components with a Conan file. This can give a green CodeQL result for code that was not built. It also leaves the webserver module uncovered after #633 removes its existing CodeQL job.

 Can we make the supported components explicit and ensure every triggered component is actually analyzed?

thompson-tomo:
> CodeQL should not require adding Conan support to each component. Native CMake or system packages are a better fit here.

Conan was used to only import dependencies hence we didn't add support for conan and they were still built using cmake.

> The workflow triggers for all exporters/ and instrumentation/ changes, but only analyzes components with a Conan file.  It also leaves the webserver module uncovered after ci #633 removes its existing CodeQL

The biggest difficulty we face is codeql requires everything to be built in the 1 workflow hence why we need a reproducible build process. The problem got even worse when you consider different components needed different versions of the same dependency as otherwise they failed. Hence the conan was added to the components which could be built with the goal being add more once they are buildable. Note prior to this pr there was no codeql  at all as the current codeql doesn't work, hence partial coverage is better than none.

lalitb:
> The biggest difficulty we face is codeql requires everything to be built in the 1 workflow hence why we need a reproducible build process. The problem got even worse when you consider different components needed different versions of the same dependency as otherwise they failed. Hence the conan was added to the components which could be built with the goal being add more once they are buildable. Note prior to this pr there was no codeql at all as the current codeql doesn't work, hence partial coverage is better than none.

I agree that partial coverage is better than none. My concern is only that the workflow should clearly represent that partial coverage.

  Right now, it triggers for every change under `exporters/**` and `instrumentation/**`, but it builds only the components that have a `conanfile.txt`. This means CodeQL can report green even when the changed component was not analyzed.

CodeQL also does not require every component to use the same build environment. Components with different dependency requirements can use separate jobs or a matrix.

Could we limit the path triggers to the components currently supported, or define those components explicitly in a matrix? We can expand that list as more components become buildable.

thompson-tomo:
> CodeQL also does not require every component to use the same build environment. Components with different dependency requirements can use separate jobs or a matrix.

I don't believe that is the case given that the build must be done between the init & analyze steps and the analyze replaces any previous analyze hence needs to be a single check.

I need to think of a maintainable & reproducible solution on how to perform codeql analysis especially if we can't use conan.

.github/workflows/codeql.yml
thompson-tomo: I don't believe package managers are classified as distributions especially as it doesn't satisfy any of the customisations checklist.

> They provide recipes/ports that change build options, patch sources, and apply ecosystem-specific toolchains/configuration files on build/install.

Yes but these are not vendor or end-user specific hence in fact it seems to fit in the category of a contribution.

thompson-tomo:
@dbarker after many, many attempts the ci is now failing after adding the fetch content rather than using conan. It has even broken the pure cmake build. 

Note The codeql error is one of the errors i saw previously when using a newer version via conan but now i get the error on all fetch content versions.

instrumentation/boost_log/CMakeLists.txt
dbarker, line 26: `OPENTELEMETRY_CPP_INCLUDE_DIRS` isn't defined when using fetch content so this results in the headers not being found. That variable can be removed from the `target_include_directories` call. 

The fix is to link `opentelemetry-cpp::api` to `${this_target}`  with `target_link_libraries` below.

dbarker (commented):
Thanks for the updates. The fetch content usage looks good. The compile errors are coming from some existing CMake issues in those files. They aren't linking the otel cpp API CMake target (this is also important to get the transitive compile definitions from the API target).

instrumentation/boost_log/CMakeLists.txt
thompson-tomo: That appears to not have worked as ci is still failing after said changes

dbarker, line 29: missing the public link to `opentelemetry-cpp::api`

```cmake
target_link_libraries(${this_target} PUBLIC opentelemetry-cpp::api Boost::log)
```

instrumentation/glog/CMakeLists.txt
```diff
@@ -7,25 +7,40 @@ set(this_target opentelemetry_glog_sink)
...
 
 set_target_properties(${this_target} PROPERTIES EXPORT_NAME ${this_target})
 
 target_include_directories(${this_target} PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-  ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
 )
 
 target_link_libraries(${this_target} PRIVATE
```
dbarker, line 29: the glog sink exposes glog and opentelemetry-cpp headers publicly in its header. Both must be linked with `PUBLIC` scope.

```cmake
target_link_libraries(${this_target} PUBLIC glog::glog opentelemetry-cpp::api)
```

instrumentation/log4cxx/CMakeLists.txt
dbarker, line 44: missing the link to opentelemetry-cpp::api. Links to the api and log4cxx must be with PUBLIC scope since the appender exposes headers for both publicly.

dbarker, line 41: ${OPENTELEMETRY_CPP_INCLUDE_DIRS} can be removed once the api is properly linked. See below.

instrumentation/boost_log/apt-packages.txt
```diff
@@ -0,0 +1,10 @@
+build-essential
```
dbarker, line 1: Please remove the `apt-packages.txt` files from this PR. 

The separate apt package files are not necessary for the purpose of running the CodeQL analysis, will introduce a maintenance burden, imply support for a package manager, and are unusual/unexpected for a C++ CMake project generally.  

All dependencies needed to support analyzing the contrib components with CodeQL can be installed directly in one job step within the scope of the CodeQL workflow file. This will keep the PR well scoped and focused on enabling that CI workflow. 

Please note several dependencies (ssl, curl, protobuf) may not be required by the CodeQL workflow and components tested.

instrumentation/log4cxx/CMakeLists.txt
```diff
@@ -7,27 +7,28 @@ set(this_target opentelemetry_log4cxx_appender)
...
-  message(FATAL_ERROR "Could not find opentelemetry-cpp")
-endif()
+FetchContent_Declare(
+  opentelemetry-cpp
+  GIT_REPOSITORY https://github.com/open-telemetry/opentelemetry-cpp.git
+  GIT_TAG 6f65c3eded2bb4656c8d675cad43bbe694c531c7 # v1.28.0
+)
 
-find_package(log4cxx QUIET)
+FetchContent_Declare(
```
dbarker: The install error below indicates that either the CMake option to install log4cxx is not enabeld or the log4cxx CMake package does not support installation when built as a subproject. 

The simple fix is to continue to install the package in the CI workflow files with apt and call find_package for log4cxx in this file. 

```sh
CMake Error in CMakeLists.txt:
  export called with target "opentelemetry_log4cxx_appender" which requires
  target "log4cxx" that is not in any export set.


CMake Error in CMakeLists.txt:
  export called with target "opentelemetry_log4cxx_appender" which requires
  target "log4cxx" that is not in any export set.


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

dbarker (commented):
Please see feedback on the latest changes below.

instrumentation/boost_log/CMakeLists.txt
dbarker: I've reviewed the errors and shared feedback in the follow-up review. Please see https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/634#pullrequestreview-4917361281

instrumentation/boost_log/apt-packages.txt
thompson-tomo: This is purely a text file which is feed to apt-get so that each instrumentation/exporter has what it needs. Hence it doesn’t imply support for a package manager. Without it you are duplicating manually managed lists of what to install including hardcoding it in ci workflows. hence it is actually providing a pathway to reducing maintenance and that same time it serves as a marker of a package which can be built in a standardised manner.

The dependencies was based on what the testing workflows defined. My intention was to review them once the ci is building.

thompson-tomo:
It is fair to say. i underestimated, the state of the repo and it's ability for the packages to follow a common build process to ensure a maintainable solution can be achieved so that codeql scanning can be performed. As such i have switched this to draft and instead raised #651 #652 & #653 to support this pr. The difference those pr's are focused on addressing the ci workflow and once implemented this pr should be automatically descoped as the issues attempting to be resolved here have already been tackled.

lalitb:
Thanks for working on dedicated CodeQL coverage - I agree with the goal. I also appreciate that some of the dependency and package-list work has already been separated into #651, #652, and #653.

Could we simplify this PR further by keeping the component CMake files unchanged and making CodeQL build them through their normal standalone interfaces?

One possible flow would be:

  1. Build and install one supported `opentelemetry-cpp` version.
  2. Use an explicit matrix of contrib components.
  3. Configure each component with the `opentelemetry-cpp` installation in `CMAKE_PREFIX_PATH`.
  4. Run CodeQL around each normal component build.

On the concern that separate `analyze` steps replace earlier results: each matrix job can pass a distinct `category` to `github/codeql-action/analyze`. That keeps the component results side by side instead of overwriting them. This lets components use different dependency environments without requiring every component to build in one aggregate CMake project.

The current aggregate build is not working yet: `boostlog` does not match the actual `boost_log` directory, and spdlog uses `find_package(opentelemetry-cpp REQUIRED)` even though the fetched source is not installed as a discoverable package. Its next `find_package(spdlog REQUIRED)` also has no provider because spdlog is neither installed by `apt-packages.txt` nor fetched. The current run therefore uploads a configuration-error SARIF rather than completing a CodeQL analysis.

The workflow also triggers for all `exporters/**` and `instrumentation/**` changes while attempting only three instrumentation components. We should either build every component the check claims to cover or narrow its path triggers.

For nginx and the webserver module, I suggest keeping the existing CodeQL job until a dedicated replacement is green and actually compiles those sources.

PR #638, opened 2026-07-26
ci: Update/cleanup CI for httpd

thompson-tomo:
This resolves runner image not available issue & removes un-necessary steps which have been promoted to shared step.


marcalff:
@thompson-tomo 

Thanks for the fix.

So, CI is failing during the build, and this is actually **progress**, because at least CI now runs on available platforms instead of not at all.

Ok to remove the per contribution format scripts and CI, this needs to be shared, to resolve with #637

marcalff (approved):
LGTM

PR #640, opened 2026-08-02
chore: Add Renovate configuration

thompson-tomo:
Added package rules for dependency management in Renovate configuration.

Note to help adoption current config will only raise pr’s automatically for github actions, the rest will be visible via dashboard.

C++ is 1 of the 2 language sigs which haven’t switched.

Renovate enables pinning of gh actions to sha, better package registry support & sourcing of packages.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,32 @@
+{
+  $schema: "https://docs.renovatebot.com/renovate-schema.json",
+  extends: [
+    "config:best-practices",
+    "helpers:pinGitHubActionDigestsToSemver",
+    ":separateMultipleMajorReleases",
+  ],
+  packageRules: [
+    {
+      dependencyDashboardApproval: true,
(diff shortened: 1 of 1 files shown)


marcalff:
@thompson-tomo 

> C++ is 1 of the 2 language sigs which haven’t switched.

Could you elaborate on this (links), I was not aware we need to do something.

thompson-tomo:
See https://github.com/open-telemetry/sig-security/issues/91

The otel repos are shown at https://developer.mend.io/github/open-telemetry with python currently rolling out atm, leaving cpp & rust.

Renovate can automatically bump the github runners hence workflows wouldn't go un-noticed that they needed updates as they are stuck on old versions like we had.

It also offers broader range of updates that can be performed including adding custom sources if needed.

Another thing it does is enforce pinning of github actions to sha as currently some are pinned but many are not.

marcalff (approved):

PR #641, opened 2026-08-03
cI: Add more renovate rules based on having done a scan.

thompson-tomo:
This analyses the renovate scan to implement rules to help mantain the dependencies going forward.

A balance between grouping and individual pr's will be raised to minimise updates blocked due to failing ci.

Each rule has a description of what it does.

This is a Continuation of https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/640 which enabled the scanning.


lalitb:
Thanks for working on this. I think Renovate is useful for this repo, but this change enables too much at once.

  Can we keep dashboard approval as the default and enable each dependency type separately after reviewing what Renovate finds?

  I also suggest keeping digest updates separate. Combining Docker, GitHub Actions, Bazel, and library updates in one PR could make CI failures difficult to understand and may block unrelated updates.

  A safer approach would be:

  - Keep automatic updates for GitHub Actions.
  - Keep other updates behind dashboard approval for now.
  - Add separate rules for Docker and library dependencies.
  - Group only closely related dependencies.

  This will give us the benefits of Renovate without creating large mixed update PRs.

thompson-tomo:
> Can we keep dashboard approval as the default and enable each dependency type separately after reviewing what Renovate finds?

Thats exactly what i did, renovate didn't find as much as i was hoping for but thats due to using apt packages, git repo's.

> I also suggest keeping digest updates separate. Combining Docker, GitHub Actions, Bazel, and library updates in one PR could make CI failures difficult to understand and may block unrelated updates.

Only patches/digest are combined everything else is seperate. If a patch breaks ci, then something is wrong as that shouldn't be the case.

> Keep other updates behind dashboard approval for now.

Have just put more behind approval.

> Add separate rules for Docker and library dependencies.

Already the case except for patches.

> Group only closely related dependencies.

Most updates are already seperate, key thing i like to avoid is having a too noisy setup especially when review bandwidth is low. Shortly after the config pr is merged the issue will show you its update plan.

thompson-tomo:
@lalitb the dashboard is available as https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/672 and even with these rules a very significant number of updates will still be behind approval (all majors) and in fact most updates will still be single updates as they are minor. Let me know if there is any further feedback after seeing the dashboard & the pool of updates.

marcalff (approved):

PR #642, opened 2026-08-03
chore: Pin docker/runners to verbose pins.

thompson-tomo:
This explicitly pins docker/runners to verbose versions which renovate will update going forward including setting the sha for docker.


exporters/user_events/Docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-FROM ubuntu:latest
+FROM ubuntu:26.10
```
lalitb: Should this be 24.04 ?

thompson-tomo: The digest are as follows:

- latest: sha256:3131b4cc82a783df6c9df078f86e01819a13594b865c2cad47bd1bca2b7063bb
- 26.10: sha256:694b773ee7e0d0b55ca74c095ac3309055589e7cbfaf3100f0b226c38c6936fa
- 26.04: sha256:3131b4cc82a783df6c9df078f86e01819a13594b865c2cad47bd1bca2b7063bb
- 24.04: sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90

Which means it currently is 26.04 so will update it.

lalitb (approved):

marcalff (approved):

Issue #322, opened 2023-08-30
ApacheModuleOtelExporterHeaders Syntax Error


khpeet:
Getting the following syntax error when trying to configure export directly in apache module conf:

```
AH00526: Syntax error on line 17 of /etc/apache2/conf-enabled/opentelemetry_module.conf:
Invalid command 'ApacheModuleOtelExporterHeaders', perhaps misspelled or defined by a module not included in the server configuration
```

Config:

```
ApacheModuleOtelExporterEndpoint https://otlp.test.net:4317
ApacheModuleOtelExporterHeaders api-key=4id9f8...
```

**Steps to reproduce**
1. Uncomment export headers
2. Try to restart apache (or stop/start)

**What is the expected behavior?**
Ability to export successfully, with no errors.

pnvnd:
I tried building `opentelemetry-webserver-sdk-x64-linux.tgz` over the weekend myself and can confirm the `ApacheModuleOtelExporterHeaders` is also not working for me.

PR #240, opened 2022-11-18
Apache and Ngnix OtelExporterOtlpHeaders changes 

gnm444:
Apache and Ngnix OtelExporterHeaders changes related to https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/198

I have tested this using Wireshark. Please find below the headers information captured using tcpdump

```
..$.....................@.....@...... .................?..
04:38:22.473273 IP 172.18.0.4.4317 > 172.18.0.2.49424: tcp 0
E..4..@.@.?..............z)/.._y....XQ.....
..<Mo. .
04:38:22.480960 IP 172.18.0.4.4317 > 172.18.0.2.49424: tcp 15
E..C..@.@.?..............z)/.._y....X`.....
..<Uo. ..............@.
04:38:22.481068 IP 172.18.0.2.49424 > 172.18.0.4.4317: tcp 0
E..4..@.@.3..............._y.z)>....XQ.....
o. ...<U
04:38:22.482460 IP 172.18.0.2.49424 > 172.18.0.4.4317: tcp 1357
E.....@.@..o.............._y.z)>....]......
o. ...<U...........5........@
:authority.collector:4317@.:path;/opentelemetry.proto.collector.trace.v1.TraceService/Export..**api-key.abc123**@.te.trailers@.content-type.application/grpc@
user-agent-grpc-c++/1.36.4 grpc-c/15.0.0 (linux; chttp2)@.grpc-accept-encoding.identity,deflate,gzip@.accept-encoding.identity,gzip@.grpc-timeout.9970m...........................
..
..
 
.telemetry.sdk.version..
.1.2.0
.
.telemetry.sdk.language..
.cpp
+
.service.namespace..
.DemoServiceNamespace
.
.service.name.
.DemoService
%
.telemetry.sdk.name..
.opentelemetry
'
.service.instance.id..
.DemoInstanceId
.
.webengine.name..
.Apache...
```

PR #645, opened 2026-08-04
fix: missing field in merged config

proost:
I digged out [this issue](https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/322), But i can't reproduce it.

But i found why [not working](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/240#issuecomment-1739892780).

We missed "otelExporterOtlpHeaders" in the merged config. so works like nothing to configured.

And in the docker file, URL is wrong. So i fixed too.


.github/workflows/webserver.yml
```diff
@@ -196,31 +196,3 @@ jobs:
       #     sleep 30
       #     ./gradlew :test:integration:integrationTests -i
       #     curl http://localhost:9411/api/v2/spans?serviceName=demoservice
-
-
-  Codeql-build:
```
lalitb, line 215: This removes the webserver CodeQL job, but the PR description does not explain the loss of static-analysis coverage, and the proposed replacement in #634 is still a failing draft.

Could we keep this job for now, or split its removal into the PR that introduces a working replacement?

PR #651, opened 2026-08-13
Define apt packages in a file

thompson-tomo:
This improves maintainability by reducing content which is defined in ci workflows which brings the benefit that they can be installed in any workflow they are applicable for.

There is no change in what is installed and the only functional difference is the packages are being passed via stdin rather as parameters on the cli command.

This also makeş it easier to in the future pin the versions.

### DIFF
--- a/exporters/fluentd/apt-packages.txt
+++ b/exporters/fluentd/apt-packages.txt
@@ -3,6 +3,3 @@ cmake
 ninja-build
 libssl-dev
 libcurl4-openssl-dev
-libgmock-dev
-libgtest-dev
-libbenchmark-dev
--- a/exporters/fluentd/apt-packages.txt
+++ b/exporters/fluentd/apt-packages.txt
@@ -1,0 +1,8 @@
+build-essential
+cmake
+ninja-build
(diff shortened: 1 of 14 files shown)


exporters/fluentd/apt-packages.txt
```diff
@@ -0,0 +1,8 @@
+build-essential
+cmake
+ninja-build
+libssl-dev
+libcurl4-openssl-dev
+libgmock-dev
+libgtest-dev
+libbenchmark-dev
```
lalitb: nit - I think `libbenchmark-dev` was not installed earlier - Is it needed?

lalitb (approved):
Thanks.

exporters/fluentd/apt-packages.txt
thompson-tomo: Was previously there but removed as part of https://github.com/open-telemetry/opentelemetry-cpp-contrib/commit/c26b8d13090f83cd41fba71076179ccb8d65441d hence will remove it.

marcalff (approved):
LGTM

PR #652, opened 2026-08-13
Move google test to a fetch content

thompson-tomo:
This moves google test into cmake for the package rather than a dedicated checkout, this means all cmake benefit rather than just ci and increases testability.

### DIFF
--- a/instrumentation/glog/CMakeLists.txt
+++ b/instrumentation/glog/CMakeLists.txt
@@ -69,7 +69,15 @@ if(OPENTELEMETRY_INSTALL)
 endif() # OPENTELEMETRY_INSTALL
 
 if(BUILD_TESTING)
-  find_package(GTest 1.12 REQUIRED)
+  include(FetchContent)
+
+  FetchContent_Declare(
+    GTest
+    GIT_REPOSITORY https://github.com/google/googletest.git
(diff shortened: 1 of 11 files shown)


instrumentation/glog/CMakeLists.txt
```diff
@@ -69,7 +71,15 @@ if(OPENTELEMETRY_INSTALL)
 endif() # OPENTELEMETRY_INSTALL
 
 if(BUILD_TESTING)
-  find_package(GTest 1.12 REQUIRED)
+  include(FetchContent)
+
+  FetchContent_Declare(
```
dbarker, line 76: Please declare with the name `googletest` here. CMake will deduplicate fetches based on this name. If a parent project wants to define their own version of googletest they can override all sub-projects' declarations if all use the same name.  

CMake uses `googletest` in their documentation for FetchContent and the otel-cpp declaration follows this too (https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/googletest.cmake#L21).

dbarker (commented):
Thanks! This is great. Please see feedback below.

instrumentation/glog/CMakeLists.txt
thompson-tomo: Done

dbarker (approved):
Looks good. Thanks!

PR #653, opened 2026-08-13
Use fetch content for otel cpp

thompson-tomo:
This goes through and removes the otel cpp repo checkout and local build steps and instead leverage a fetchcontent in the cmake build process.

It is easier to review this pr hiding whitespace aka https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/653/changes?w=1#diff-afedc88ba9243d94695a1f609b7a87dab38fc0dd1a86a789eaba4ddffe4780bf


thompson-tomo:
Looking for input as to what to do to resolve ci failures.

exporters/geneva-trace/CMakeLists.txt
```diff
@@ -30,6 +36,10 @@ set_target_properties(
   opentelemetry_exporter_geneva_logs
   PROPERTIES EXPORT_NAME opentelemetry_exporter_geneva_logs)
 
+if(MAIN_PROJECT)
+  target_link_libraries(opentelemetry_exporter_geneva_logs
```
dbarker: Both the FetchContent and find_package paths import the public `opentelemetry-cpp::*` CMake targets and the call to `target_link_libraries( ... opentelemetry-cpp::api)` should be unconditional.

instrumentation/boost_log/src/sink.cc
```diff
@@ -94,17 +94,17 @@ OpenTelemetrySinkBackend::OpenTelemetrySinkBackend(const ValueMappers &mappers)
 
   set_file_path_if_valid_ = {[](LogRecord *, const std::string &) {},
                              [](LogRecord *log_record, const std::string &file_name) {
-                               log_record->SetAttribute(kCodeFilepath, file_name);
+                               log_record->SetAttribute("code.filepath", file_name);
```
dbarker, line 97: Please use the auto generated code attribute constants from `<opentelemetry/semconv/code_attributes.h>`

instrumentation/boost_log/CMakeLists.txt
```diff
@@ -5,69 +5,73 @@ cmake_minimum_required(VERSION 3.16)
...
 set_target_properties(${this_target} PROPERTIES EXPORT_NAME ${this_target})
 
 target_include_directories(${this_target} PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-  ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
 )
 
+if(MAIN_PROJECT)
+  target_link_libraries(${this_target} PUBLIC
```
dbarker: The link here should be unconditional

instrumentation/boost_log/src/sink.cc
thompson-tomo: I can't these are the deprecated constants which no longer exist and, changing it to the new constant would be a breaking change.

dbarker: I missed that these are the deprecated attributes. 

The opentelemetry-cpp api target includes the `<opentelemetry/semconv/incubating/code_attributes.h>` [header](https://github.com/open-telemetry/opentelemetry-cpp/blob/f77c1a5cae46659b4666636d45dfeb6056d320c0/api/include/opentelemetry/semconv/incubating/code_attributes.h#L54) with the deprecated versions. Can these be used?

thompson-tomo: Thanks, i had missed that the deprecated were still in the incubating conventions. Have switched them but i am more interested in what is needed to fix ci after making the previously requested reverts.

instrumentation/boost_log/CMakeLists.txt
dbarker:
```sh
CMake Error in CMakeLists.txt:
  export called with target "opentelemetry_boost_log_sink" which requires
  target "opentelemetry_api" that is not in any export set.
```

This export method is causing the configuration errors in CI. Please remove it here and the other instances. 

It is disabled by default since v3.15 by CMP0090 and we shouldn't support it.
https://cmake.org/cmake/help/latest/policy/CMP0090.html

thompson-tomo: Thanks @dbarker that managed to resolve most of the failing ci & i was able to resolve the remaining failing ci.

Please take another look.

PR #658, opened 2026-08-19
Switch tracepoint to fetch

thompson-tomo:
This switches tracepoint from gitmodule to a fetch content.


exporters/user_events/CMakeLists.txt
```diff
@@ -6,7 +6,6 @@ endif()
 
 option(WITH_EXAMPLES "Build example" ON)
 option(BUILD_TESTING "Build tests" ON)
-option(BUILD_TRACEPOINTS "Build tracepoints library" ON)
```
dbarker, line 9: The BUILD_TRACEPOINTS option is referenced explicitly in the vcpkg port and appears to be set to OFF when building with the `user_events` feature ON. 

https://github.com/microsoft/vcpkg/blob/06d00ffa491e4668627728f14b891d22c6fea146/ports/opentelemetry-cpp/portfile.cmake#L29

@ThomsonTan should tracepoints remain optional?

dbarker (commented):
Thanks! Removing the submodule and switch to FetchContent look good. 

Only concern with this change is removal of the BUILD_TRACEPOINTS option since it is referenced by the vcpkg port (the main consumer of this contrib component).

Safer to keep the option for this PR and follow up if needed on removing it.

thompson-tomo:
I thought that option was used only in the section that was removed hence why i removed it.

exporters/user_events/CMakeLists.txt
lalitb: Did a quick check on the current [vcpkg port](https://github.com/microsoft/vcpkg/blob/master/ports/opentelemetry-cpp/portfile.cmake), and it still sets `BUILD_TRACEPOINTS=OFF` for the `user-events` feature.

  So we should keep this option. We can use FetchContent when it is `ON` and continue using the installed tracepoint packages when it is `OFF`.

thompson-tomo: Been changed

dbarker (approved):
Thanks for the updates. Sorry for the delay in review.

exporters/user_events/CMakeLists.txt
```diff
@@ -26,20 +26,28 @@ if(MAIN_PROJECT)
   find_package(opentelemetry-cpp REQUIRED)
 endif()
 
-if(BUILD_TRACEPOINTS)
+if(NOT BUILD_TRACEPOINTS)
+  find_package(eventheader-headers QUIET)
+endif()
+
+if(BUILD_TRACEPOINTS OR NOT eventheader_headers_FOUND)
```
ThomsonTan: Should the variable be eventheader-headers_FOUND? Set by the `eventheader-headers` package.

exporters/user_events/CMakeLists.txt
```diff
@@ -26,20 +26,28 @@ if(MAIN_PROJECT)
   find_package(opentelemetry-cpp REQUIRED)
 endif()
 
-if(BUILD_TRACEPOINTS)
+if(NOT BUILD_TRACEPOINTS)
+  find_package(eventheader-headers QUIET)
```
ThomsonTan: Even after correcting the below FOUND-variable typo, OFF with a missing package still falls back to building the dependency, whereas the previous REQUIRED lookup stopped configuration. That changes the installed-package contract rather than merely changing how source dependencies are obtained. Please keep an exclusive OFF branch with `find_package(eventheader-headers REQUIRED CONFIG)` and test that missing packages fail before tracepoints population.

ThomsonTan: This lookup also drops CONFIG, changing the previous config-only package resolution. Probably add it back?

thompson-tomo: It all comes down to what someone's impression of what the build option should do is, for me it was ignore what maybe installed and instead build it. Hence using the fetch if not present was an improvement.

Anyway have descoped pr just to get it in.

dbarker: The recent changes add back the  `find_package(eventheader-headers REQUIRED CONFIG)` path when BUILD_TRACEPOINTS is OFF.

exporters/user_events/CMakeLists.txt
dbarker: This change was reverted. The original condition on BUILD_TRACEPOINTS is used.

dbarker (approved):
Thanks for the updates. The scope of the PR is now just to remove the LinuxTracepoints git submodule and replace the direct add_subdirectory subproject build with FetchContent.

PR #659, opened 2026-08-19
Replace find package in glog, geneva & fleuntd

thompson-tomo:
This now has fleuntd, geneva & glog obtaining dependencies via fetchContent rather than git repo checkout or package install.

Fleuntd now also does a main project testing

### DIFF
--- a/exporters/fluentd/CMakeLists.txt
+++ b/exporters/fluentd/CMakeLists.txt
@@ -55,11 +55,6 @@ else()
   endif()
 endif()
 
-if(MAIN_PROJECT)
-  find_package(CURL REQUIRED)
-  find_package(Threads REQUIRED)
-endif()
-
 include_directories(include)
 
 if(OPENTELEMETRY_ENABLE_FLUENT_RESOURCE_PUBLISH_EXPERIMENTAL)
--- a/exporters/geneva/CMakeLists.txt
+++ b/exporters/geneva/CMakeLists.txt
@@ -74,6 +74,8 @@ if(BUILD_TESTING)
 
   FetchContent_MakeAvailable(googletest)
 
+  include_directories(SYSTEM test/decoder)
+
   enable_testing()
 
   set(testname geneva_metrics_exporter_test)
@@ -83,8 +85,13 @@ if(BUILD_TESTING)
   # build geneva metrics tests
   add_compile_definitions(KS_STR_ENCODING_NONE)
   add_executable(${testname}
-    test/metrics_exporter_test.cc test/decoder/ifx_metrics_bin.cpp
+    test/metrics_exporter_test.cc
+    test/decoder/ifx_metrics_bin.cpp
     test/decoder/kaitai/kaitaistream.cpp)
+
+  target_include_directories(${testname} PRIVATE
+    ${CMAKE_CURRENT_LIST_DIR}/test/decoder/kaitai
+  )
 
   target_include_directories(${testname} PUBLIC
     ${CMAKE_CURRENT_LIST_DIR}/include
@@ -94,12 +101,12 @@ if(BUILD_TESTING)
   target_link_libraries(${testname} PRIVATE
     GTest::gmock
     GTest::gtest
+    GTest::gtest_main
     opentelemetry_exporter_geneva_metrics)
 
   gtest_add_tests(
     TARGET ${testname}
     TEST_PREFIX exporter.
-    TEST_LIST ${testname}
   )
 endif()
 
--- a/exporters/geneva/CMakeLists.txt
+++ b/exporters/geneva/CMakeLists.txt
@@ -1,9 +1,14 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
 cmake_minimum_required(VERSION 3.12)
 
 # MAIN_PROJECT CHECK determine if fluentd exporter is built as a subproject
 # (using add_subdirectory) or if it is the main project
 #
 set(MAIN_PROJECT OFF)
+
+include(FetchContent)
 
 if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
   project(opentelemetry-geneva-metrics)
@@ -61,37 +66,41 @@ endif()
 set_target_version(opentelemetry_exporter_geneva_metrics)
 
 if(BUILD_TESTING)
-  if(EXISTS ${CMAKE_BINARY_DIR}/lib/libgtest.a)
-    # Prefer GTest from build tree. GTest is not always working with
-    # CMAKE_PREFIX_PATH
-    set(GTEST_INCLUDE_DIRS
-        ${CMAKE_CURRENT_SOURCE_DIR}/third_party/googletest/googletest/include
-        ${CMAKE_CURRENT_SOURCE_DIR}/third_party/googletest/googlemock/include)
-    set(GTEST_BOTH_LIBRARIES
-        ${CMAKE_BINARY_DIR}/lib/libgtest.a
-        ${CMAKE_BINARY_DIR}/lib/libgtest_main.a
-        ${CMAKE_BINARY_DIR}/lib/libgmock.a)
-  else()
-    find_package(GTest REQUIRED)
-  endif()
-  include_directories(SYSTEM ${GTEST_INCLUDE_DIRS})
-  include_directories(SYSTEM test/decoder)
+  FetchContent_Declare(
+    googletest
+    GIT_REPOSITORY https://github.com/google/googletest.git
+    GIT_TAG 15460959cbbfa20e66ef0b5ab497367e47fc0a04 # v1.12.0
+  )
+
+  FetchContent_MakeAvailable(googletest)
+
   enable_testing()
+
+  set(testname geneva_metrics_exporter_test)
+
   include(GoogleTest)
+
   # build geneva metrics tests
   add_compile_definitions(KS_STR_ENCODING_NONE)
-  add_executable(
-    geneva_metrics_exporter_test
+  add_executable(${testname}
     test/metrics_exporter_test.cc test/decoder/ifx_metrics_bin.cpp
     test/decoder/kaitai/kaitaistream.cpp)
-  target_link_libraries(
-    geneva_metrics_exporter_test ${GTEST_BOTH_LIBRARIES}
-    ${CMAKE_THREAD_LIBS_INIT} opentelemetry_exporter_geneva_metrics)
+
+  target_include_directories(${testname} PUBLIC
+    ${CMAKE_CURRENT_LIST_DIR}/include
+    ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+  )
+
+  target_link_libraries(${testname} PRIVATE
+    GTest::gmock
+    GTest::gtest
+    opentelemetry_exporter_geneva_metrics)
 
   gtest_add_tests(
-    TARGET geneva_metrics_exporter_test
+    TARGET ${testname}
     TEST_PREFIX exporter.
-    TEST_LIST geneva_metrics_exporter_test)
+    TEST_LIST ${testname}
+  )
 endif()
 
 if(OPENTELEMETRY_INSTALL)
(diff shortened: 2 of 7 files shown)


instrumentation/glog/CMakeLists.txt
```diff
@@ -10,7 +10,13 @@ project(${this_target})
 include(FetchContent)
 
 find_package(opentelemetry-cpp REQUIRED)
-find_package(glog 0.7.0 REQUIRED)
+FetchContent_Declare(
```
lalitb: This makes every standalone glog instrumentation build download and build its own copy of `glog`, even when the consumer already provides a compatible installed package. That breaks offline package builds and prevents the parent project from controlling the dependency.

Could we keep `find_package(glog 0.7.0 QUIET)` as the first choice and use this pinned FetchContent declaration only as a fallback or opt-in?

thompson-tomo: Been changed

instrumentation/glog/CMakeLists.txt
```diff
@@ -10,7 +10,17 @@ project(${this_target})
 include(FetchContent)
 
 find_package(opentelemetry-cpp REQUIRED)
-find_package(glog 0.7.0 REQUIRED)
+find_package(glog QUIET)
```
lalitb: Thanks, this is the right fallback structure. It looks like the `0.7.0` minimum from the original lookup was dropped, though. Could we keep it as:

```cmake
  find_package(glog 0.7.0 QUIET)
```

exporters/geneva/CMakeLists.txt
```diff
@@ -61,37 +66,48 @@ endif()
...
-    find_package(GTest REQUIRED)
-  endif()
-  include_directories(SYSTEM ${GTEST_INCLUDE_DIRS})
+  FetchContent_Declare(
+    googletest
+    GIT_REPOSITORY https://github.com/google/googletest.git
+    GIT_TAG 15460959cbbfa20e66ef0b5ab497367e47fc0a04 # v1.12.0
+  )
+
+  FetchContent_MakeAvailable(googletest)
```
lalitb, line 75: I believe we need to bump cmake version to use this - https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable

dbarker: Good catch. We need to upgrade all the CMake files in contrib to 3.16 for minimum. It shouldn't be blocking since the github runners package newer CMake versions.

thompson-tomo: Done via https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/669 with the scope limited to packages with working ci.

exporters/geneva/CMakeLists.txt
```diff
@@ -61,37 +66,48 @@ endif()
...
-    geneva_metrics_exporter_test ${GTEST_BOTH_LIBRARIES}
-    ${CMAKE_THREAD_LIBS_INIT} opentelemetry_exporter_geneva_metrics)
+
+  target_include_directories(${testname} PRIVATE
+    ${CMAKE_CURRENT_LIST_DIR}/test/decoder/kaitai
+  )
+
+  target_include_directories(${testname} PUBLIC
+    ${CMAKE_CURRENT_LIST_DIR}/include
+    ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
```
dbarker: The `OPENTELEMETRY_CPP_INCLUDE_DIRS` provides support for legacy CMake style includes and can be removed here. Instead we should prefer linking the geneva targets to the required `opentelemetry-cpp` targets in the `target_link_libraries` call below.

exporters/geneva/CMakeLists.txt
```diff
@@ -125,9 +141,19 @@ if(OPENTELEMETRY_INSTALL)
...
-  add_executable(example_metrics example/example_metrics.cc
+  set(example_exe example_metrics)
+  add_executable(${example_exe} example/example_metrics.cc
                                  example/foo_library.cc)
-  target_link_libraries(example_metrics opentelemetry_exporter_geneva_metrics)
+  set_target_properties(${example_exe} PROPERTIES EXPORT_NAME ${example_exe})
+
+  target_include_directories(${example_exe} PUBLIC
+    ${CMAKE_CURRENT_LIST_DIR}/include
+    ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
```
dbarker: Same as above. Please link the required opentelemetry-cpp targets below to allow removal of the OPENTELEMETRY_CPP_INCLUDE_DIRS include here.

thompson-tomo: Have made changes. Could you please take another look.

exporters/geneva/CMakeLists.txt
thompson-tomo: Have made changes. Could you please take another look.

dbarker: Thanks. We need to determine what opentelemetry-cpp targets this component actually needs to prevent linking all targets installed (OPENTELEMETRY_CPP_LIBRARIES includes every target installed with the opentelemetry-cpp package). This can help us narrow down what external dependencies this component has too.

thompson-tomo: Would it be possible to handle the otel-cpp changes as part of https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/653 that way we can also make it consistent across multiple.

dbarker: Lets fix the root cause of the example and test linking problems in this PR. 

The issue is this conditional: 

```cmake
if(MAIN_PROJECT)
   target_include_directories(opentelemetry_exporter_geneva_metrics
                      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS})
   target_link_libraries(opentelemetry_exporter_geneva_metrics
                      PUBLIC ${OPENTELEMETRY_CPP_LIBRARIES})
else()
   target_link_libraries(opentelemetry_exporter_geneva_metrics
                      PUBLIC opentelemetry_metrics opentelemetry_resources opentelemetry_common)
endif()
```

The problem that the geneva metrics target links libraries differently if the project is a main project and does not use the imported targets for otel-cpp. When MAIN_PROJECT is true it falls back to legacy style include directories as PRIVATE and so the tests and examples don't find the otel-cpp headers.

The fix is to publicly link the otel-cpp imported targets in both cases: 

```cmake
   target_link_libraries(opentelemetry_exporter_geneva_metrics
                      PUBLIC opentelemetry-cpp::metrics opentelemetry-cpp::resources opentelemetry-cpp::common)
````

Comment (thompson-tomo) [reply exporters/geneva/CMakeLists.txt]: Done

Comment (dbarker) [reply exporters/geneva/CMakeLists.txt]: Thanks. Since the imported targets include the directories needed we can remove the conditional and  `target_include_directories(opentelemetry_exporter_geneva_metrics PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS})`

File: exporters/geneva/CMakeLists.txt [hunk 6]
```diff
@@ -125,9 +134,12 @@ if(OPENTELEMETRY_INSTALL)
 endif()
 
 if(WITH_EXAMPLES)
-  add_executable(example_metrics example/example_metrics.cc
+  set(example_exe example_metrics)
```
dbarker, line 132: why the extra indirection with a local variable?

exporters/geneva/CMakeLists.txt
```diff
@@ -125,9 +134,12 @@ if(OPENTELEMETRY_INSTALL)
...
+  add_executable(${example_exe} example/example_metrics.cc
                                  example/foo_library.cc)
-  target_link_libraries(example_metrics opentelemetry_exporter_geneva_metrics)
+  set_target_properties(${example_exe} PROPERTIES EXPORT_NAME ${example_exe})
```
dbarker: the example shouldn't be installed so no need for an export name

thompson-tomo: Done.

exporters/geneva/CMakeLists.txt
thompson-tomo: Consistency and a number of packages do it this way to minimise chance of errors etc.

dbarker: Definitely a personal preference and not blocking. I find it makes these files harder to read. CMake will tell us one way or another if the target name has a typo :)

dbarker (approved):
Thanks for the build fixes.

exporters/fluentd/CMakeLists.txt
```diff
@@ -55,11 +55,6 @@ else()
   endif()
 endif()
 
-if(MAIN_PROJECT)
-  find_package(CURL REQUIRED)
-  find_package(Threads REQUIRED)
```
proost, line 60: If this exporter uses as main, we need Threads. can you double-check?

thompson-tomo: Isn't that what the ci is testing?

proost: If that is true, then should leave logs into stdout but i can't see them. And it looks like depending on the "opentelemetry-cpp" build.

thompson-tomo: You are right fleuntd ci was different to the others which do test as a main project. Have added additional ci steps.

thompson-tomo: @proost please take a look now, with additional ci checks it is still passing.

proost: Thanks!

proost (approved):
LGTM

instrumentation/glog/CMakeLists.txt
```diff
@@ -23,11 +33,12 @@ set_target_properties(${this_target} PROPERTIES EXPORT_NAME ${this_target})
...
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-  ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
 )
 
-target_link_libraries(${this_target} PRIVATE
+target_link_libraries(${this_target} PUBLIC
   glog::glog
+  opentelemetry-cpp::logs
+  opentelemetry-cpp::ostream_log_record_exporter
```
dbarker: Please only link to the `opentelemetry-cpp::api` header only target here. 

Since the main glog project library is instrumentation, it should only have a dependency on the OTel API.  The two targets linked here (`logs`, `ostream_log_record_exporter`) bring in SDK dependencies and that isn't always clear from the names.

thompson-tomo: Done

instrumentation/glog/CMakeLists.txt
```diff
@@ -10,7 +10,17 @@ project(${this_target})
...
+
+if(NOT glog_FOUND)
+  FetchContent_Declare(
+    glog
+    GIT_REPOSITORY https://github.com/google/glog.git
+    GIT_TAG 34b8da6496aec6a98277808701cfa834fae9801f # v0.7.0
+  )
+
+  FetchContent_MakeAvailable(glog)
+endif()
```
dbarker, line 24: Since the fetch content path is now an option to import `glog::glog` and the `find_package` path will never fail we need a check that the required imported target exists in order to fail fast at CMake configure time instead of at build time. 

```cmake
if(NOT TARGET glog::glog)
  message(FATAL_ERROR "The required glog::glog target was not imported")
endif()
```

This can be a follow up PR and the same pattern should be updated elsewhere.

instrumentation/glog/CMakeLists.txt
```diff
@@ -23,11 +33,11 @@ set_target_properties(${this_target} PROPERTIES EXPORT_NAME ${this_target})
 target_include_directories(${this_target} PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-  ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
 )
 
-target_link_libraries(${this_target} PRIVATE
+target_link_libraries(${this_target} PUBLIC
```
dbarker, line 38: In a follow up PR we should update these instrumented libraries to not leak the opentelemetry-cpp API dependency and make this link private. This will require C++ changes and for the log bridges that appears minimal.

PR #660, opened 2026-08-19
switch to fetch benchmark

thompson-tomo:
This switches user events to pull in google benchmark using fetch content rather than git fetch

### DIFF
--- a/.github/workflows/user_events.yml
+++ b/.github/workflows/user_events.yml
@@ -45,14 +45,10 @@ jobs:
             libprotobuf-dev \
             protobuf-compiler \
             libgmock-dev \
-            libgtest-dev \
-            libbenchmark-dev
+            libgtest-dev
 
       - name: run tests
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/user_events.yml
```diff
@@ -45,14 +46,10 @@ jobs:
             libprotobuf-dev \
             protobuf-compiler \
             libgmock-dev \
-            libgtest-dev \
-            libbenchmark-dev
+            libgtest-dev
```
dbarker, line 49: can libgtest-dev be removed also since googletest is fetched?

dbarker (approved):
Thanks! Just a non-blocking question.

.github/workflows/user_events.yml
thompson-tomo: Good point, will do that via a seperate pr as it will need to do be done for multiple packages and suspicion is other packages can also be removed.

PR #661, opened 2026-08-19
ci: Increase ci consistency in setup

thompson-tomo:
All workflows now have manual triggers.

This pr also restricts all push triggers to main to improve ci consistency and remove unnecessary checks on pr’s. To build for a commit use the manual trigger or use a draft pr.


.github/workflows/boost_log.yml
```diff
@@ -1,14 +1,16 @@
 name: boost_log
 
 on:
+  workflow_dispatch:
   push:
```
lalitb, line 5: nit - While we are normalizing these trigger blocks, `path:` should be `paths:`. GitHub otherwise applies no push path filter, so every push to `main` can run this component workflow.

The same fix is needed in `glog.yml`, `log4cxx.yml`, `spdlog.yml`, and `user_events.yml`.

lalitb (approved):
Thanks.

.github/workflows/boost_log.yml
thompson-tomo: Done

PR #665, opened 2026-08-21
ci: Source log4cxx via apt package

thompson-tomo:
This switches log4cxx to be sourced from apt package rather than source which is possible due to the package having been updated.


lalitb (approved):
Thanks.

PR #666, opened 2026-08-21
fix ci for webserver

thompson-tomo:
This bumps runner image in webserver ci & removes the broken codeql.

### DIFF
--- a/.github/workflows/webserver.yml
+++ b/.github/workflows/webserver.yml
@@ -1,13 +1,17 @@
 name: webserver instrumentation CI
 
 on:
+  workflow_dispatch:
   push:
-    branches: [ main ]
+    branches: 
+      - 'main'
     paths:
       - 'instrumentation/otel-webserver-module/**'
       - '.github/workflows/webserver.yml'
   pull_request:
-    branches: [ main, webserver-dev ]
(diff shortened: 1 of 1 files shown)


.github/workflows/webserver.yml
```diff
@@ -196,31 +197,3 @@ jobs:
       #     sleep 30
       #     ./gradlew :test:integration:integrationTests -i
       #     curl http://localhost:9411/api/v2/spans?serviceName=demoservice
-
-
-  Codeql-build:
```
lalitb, line 201: I understand the current CodeQL job cannot stay as-is because it keeps the workflow red. My concern is the sequencing: #634 is not green yet and does not currently build the webserver sources.

Could we first make #634, or another replacement, provide working webserver coverage and then remove this job? Otherwise this change fixes CI by removing the check without replacing its coverage.

thompson-tomo: The coverage appears to have never ever worked based on the comment about it now being enabled when i finally had a run succeed in my branch. This means we are not actually removing anything that has been used. The issues with this ci even after upgrading the runner image the commands don't work.

lalitb (approved):

PR #667, opened 2026-08-22
ci: Resolving nginx & prometheus build issue

thompson-tomo:
Contributes to #668

NGINX instrumentation build failures have been addressed by explicitly using buildx & switching tests to use php docker image which exists. Have also updated the tests to use nil for remote-parent-id rather than empty string.

Prometheus bazel build resolved by specifying an older bazel version to use.

Prometheus cmake build resolved by changing zlib adding process as well as switching to a native cmake build.

This is a replacement for #633 which i accidentally broke which triggered the closing of it.


exporters/prometheus/CMakeLists.txt
```diff
@@ -20,25 +20,24 @@ project(
...
-endif()
-
-if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
-  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY
-      "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}")
-endif()
+include(FetchContent)
 
-link_directories(${CMAKE_ARCHIVE_OUTPUT_DIRECTORY})
+set(ENABLE_TESTING OFF CACHE BOOL "")
```
lalitb: Can you check this once - it seems to be disabling the test for both dependencies and the exporter. Also seen this in CI:

```
  Updating test list for fixtures
  Added 0 tests to meet fixture requirements
  Checking test dependency graph...
  Checking test dependency graph end
  No tests were found!!! ??

```

exporters/prometheus/CMakeLists.txt
```diff
@@ -161,7 +167,6 @@ endif()
 
 install(
   TARGETS opentelemetry_prometheus_push_exporter
-  EXPORT "${PROJECT_NAME}-target"
```
lalitb, line 164: why is this removed ?

exporters/prometheus/CMakeLists.txt
```diff
@@ -175,33 +180,3 @@ install(
 if(BUILD_TESTING)
   add_subdirectory(test)
 endif()
-
-include(CMakePackageConfigHelpers)
-file(MAKE_DIRECTORY "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/cmake/${PROJECT_NAME}")
-
-configure_package_config_file(
```
lalitb, line 182: With this removal, the downstream CMake projects can no longer discover and link the installed exporter through its existing package target.

lalitb:
Thanks for continuing to improve the CI across contrib - these fixes are really appreciated. For this PR, the NGINX changes look good. Please revisit the Prometheus CMake workflow and dependency changes so we preserve the existing test, dependency, and packaging contracts.

thompson-tomo:
I am quite confused about the prometheus package purpose on the surface it should be an exporter just like user_events, however when you look at the cmake its name indicates otherwise.

If the name is correct then shouldn't there be a file in the root for the contrib repo & enable the prometheus to mirror other exporters. In the case of the later then the name etc should be updated.

exporters/prometheus/CMakeLists.txt
thompson-tomo: Has this been resolved with the newer commits?

exporters/prometheus/CMakeLists.txt
```diff
@@ -14,31 +14,35 @@
...
-  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY
-      "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}")
+set(MAIN_PROJECT OFF)
+
+if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  set(MAIN_PROJECT ON)
+  project(
+    opentelemetry-cpp-contrib
+    VERSION "1.19.0" # opentelemetry-cpp-contrib 1.19.0
+    HOMEPAGE_URL "https://github.com/open-telemetry/opentelemetry-cpp-contrib"
```
thompson-tomo, line 24: This doesn't look right. Shouldn't the name have prometheus?

dbarker: Good catch. This is not right. The project name should be something specific to the component.

exporters/prometheus/CMakeLists.txt
```diff
@@ -12,33 +12,37 @@
...
 endif()
 
-if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
-  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY
-      "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}")
-endif()
+find_package(ZLIB REQUIRED)
 
-link_directories(${CMAKE_ARCHIVE_OUTPUT_DIRECTORY})
+FetchContent_Declare(
```
dbarker: Please add support to find or fetch prometheus-cpp to allow building against an installed package and a check to ensure the required prometheus-cpp targets have been imported. 

This file shows the pattern and check used in otel-cpp. 

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/prometheus-cpp.cmake

thompson-tomo: Can u take a look at the recent changes & advise if the changes are as expected.

dbarker: Looks good. Thanks.

marcalff:
CI is failing since debian:11 is now EOL. Update to debian 13 ?

thompson-tomo:
@marcalff have bumped it to debian 12

exporters/prometheus/CMakeLists.txt
```diff
@@ -159,49 +179,29 @@ elseif(UNIX OR MINGW)
...
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
+  install(
+    DIRECTORY include/opentelemetry/exporters/prometheus
+    DESTINATION include/opentelemetry/exporters/
+    FILES_MATCHING
+    PATTERN "*.h")
+else()
+  otel_add_component(
+    COMPONENT exporters_prometheus
```
dbarker: This component name will conflict with the opentelemetry-cpp `exporters_prometheus` component. 

Please rename to `exporters_prometheus_push`

exporters/prometheus/CMakeLists.txt
```diff
@@ -14,31 +14,44 @@
...
-  opentelemetry-cpp-contrib
-  VERSION "1.19.0" # opentelemetry-cpp-contrib 1.19.0
-  HOMEPAGE_URL "https://github.com/open-telemetry/opentelemetry-cpp-contrib"
-  LANGUAGES CXX)
-
-if(NOT CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
-  set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY
-      "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}")
-endif()
+set(this_target opentelemetry_prometheus)
```
dbarker: Since the target is `opentelemetry_prometheus_push_exporter` (to differentiate from the otel-cpp prometheus pull exporter) please make the project name `opentelemetry-cpp-contrib-prometheus-push-exporter`

exporters/prometheus/CMakeLists.txt
dbarker: When built as a main project these lines are required to install the package. Please revert these removals for that case. 

To keep the import target name the same please update the project namespace to be `opentelemetry-cpp-contrib::`

exporters/prometheus/CMakeLists.txt
thompson-tomo: Done

exporters/prometheus/CMakeLists.txt
thompson-tomo: Done

exporters/prometheus/CMakeLists.txt
thompson-tomo: These were causing build issues when moving away from using a local version of otel cpp & also other packages didn't have this but did have the add otel_component.

My impression was that is the newer pattern?

dbarker: Can you share more about the build issues? I'm hopeful we can get the independent install working. 

The `otel_add_component` function is part of the main opentelemtry-cpp project's cmake utilities ([added in v1.21](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#using-opentelemetry-cpp-package-components)) and was used in the `user_events` and `geneva` CMake files to allow injecting those contrib components into the main opentelemetry-cpp package when creating the [vcpkg port](https://github.com/microsoft/vcpkg/blob/52d80838fb40c755b1615fbc9c7b994a33742a22/ports/opentelemetry-cpp/portfile.cmake#L52). 

I don't think we should propagate that pattern to inject external components into the core package any further in this repo. An alternative approach is to create a top level project for opentelemetry-cpp-contrib with its own utility functions to manage the install of the components. I started prototyping this in https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/550.

thompson-tomo: From memory it was related to when switching to fetching of the libraries as opposed to installing them it caused the generated *-config.cmake to fail or something like that.

> I don't think we should propagate that pattern to inject external components into the core package any further in this repo. An alternative approach is to create a top level project for opentelemetry-cpp-contrib with its own utility functions to manage the install of the components.

That sounds like a good solution as it would also enable testing of the non main project path in ci.

exporters/prometheus/CMakeLists.txt
```diff
@@ -14,31 +14,44 @@
...
+    prometheus-cpp
+    GIT_REPOSITORY https://github.com/jupp0r/prometheus-cpp.git
+    GIT_TAG e5fada43131d251e9c4786b04263ce98b6767ba5 # v1.3.0
+  )
+
+  FetchContent_MakeAvailable(prometheus-cpp)
+endif()
+
+if(NOT TARGET prometheus-cpp::core OR
+   NOT TARGET prometheus-cpp::pull)
```
proost: `prometheus-cpp::push`?

exporters/prometheus/apt-packages.txt
```diff
@@ -0,0 +1,15 @@
...
+ca-certificates
+cmake
+ninja-build
+libssl-dev
+libcurl4-openssl-dev
+libprotobuf-dev
+protobuf-compiler
+libbenchmark-dev
+libgmock-dev
+libgtest-dev
```
proost: if fetch in CMake, do we need?

thompson-tomo: Done

exporters/prometheus/CMakeLists.txt
thompson-tomo: Done

exporters/prometheus/CMakeLists.txt
dbarker: To move forward on this PR please revert the removal of the project's config file install (lines 178-207). Without those lines the project won't install a config.cmake file that is needed for users to call `find_package(opentelemetry-cpp-contrib-prometheus-push-exporter CONFIG)`

exporters/prometheus/.bazelversion
```diff
@@ -0,0 +1 @@
+7.2.0
```
marcalff: See https://bazel.build/release

Bazel 7.x will EOL in Dec 2026.

I suggest to use bazel 8.8.0, to align with opentelemetry-cpp.

marcalff (commented):
See comment on bazel versions

exporters/prometheus/.bazelversion
thompson-tomo: Done

marcalff:
@thompson-tomo 

Thanks for the cleanup.

Some remaining items for bazel:

> WARNING: /home/runner/work/opentelemetry-cpp-contrib/opentelemetry-cpp-contrib/otel_cpp_contrib/exporters/prometheus/MODULE.bazel:4:7: The attribute 'compatibility_level' in module() is a no-op and will be removed in a future Bazel release. Please remove it from your MODULE.bazel file.

Remove 'compatibility_level'.

> WARNING: For repository 'com_google_absl', the root module requires module version abseil-cpp@20240116.1, but got abseil-cpp@20260107.1 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off

Adjust to abseil-cpp@20260107.1

> WARNING: For repository 'com_google_googletest', the root module requires module version googletest@1.14.0.bcr.1, but got googletest@1.17.0.bcr.2 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off

Adjust to googletest@1.17.0.bcr.2

thompson-tomo:
@marcalff any suggestions on how to fix the failing ci which is only occurring on ubuntu but not macos?

Note i had left out those dependency bumps as renovate will raise pr's for them tomorrow.

marcalff:
> @marcalff any suggestions on how to fix the failing ci which is only occurring on ubuntu but not macos?

The failing CI is:

> FAIL: //:prometheus_push_exporter_test (Exit 127) (see /home/runner/.cache/bazel/236c0ff7721a987cf9013ff9241d0bb2/execroot/_main/bazel-out/k8-fastbuild/testlogs/prometheus_push_exporter_test/test.log)

So this is a test issue, not build.

Of course it needs to be addressed, however:

* considering the state of the current CI
* considering that this PR is needed to just BUILD first
* considering that this PR blocks the whole renovate train of upgrades needed in general

I am willing to merge build fixes first (this PR as is) and address test failures separately.

Once we get everything to build with current dependencies, we can proceed to cleanup the test suite.

@open-telemetry/cpp-contrib-maintainers Ok to merge to unblock builds in ci ?

thompson-tomo:
Thanks @marcalff I like that proposal and as such I have reverted changes I made in an attempt to fix the tests on Ubuntu.

marcalff (approved):
LGTM,

This fixes the build only to unblock CI upgrades.

Existing test failures to address separately.

exporters/prometheus/CMakeLists.txt
```diff
@@ -159,47 +179,59 @@ elseif(UNIX OR MINGW)
...
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
+  install(
+    DIRECTORY include/opentelemetry/exporters/prometheus
+    DESTINATION include/opentelemetry/exporters/
+    FILES_MATCHING
+    PATTERN "*.h")
+else()
+  otel_add_component(
```
dbarker: The `otel_add_component` pattern isn't ready to roll out to this repo as it creates a dependency on the internal CMake files from opentelemetry-cpp and is only used for a special case for the user event and genenva components to allow the vcpkg port to inject them into the opentelemetry-cpp package. 

Please remove it here and from other PRs that add it.

thompson-tomo: Done

dbarker:
>  Ok to merge to unblock builds in ci ?

How about disabling [nginx instrumentation CI / nginx (alpine:3.20, 1.27.2) (pull_request)](https://github.com/open-telemetry/opentelemetry-cpp-contrib/actions/runs/35622222916/job/106407796842?pr=667) and [prometheus instrumentation CI / Bazel on ubuntu-24.04 (pull_request)](https://github.com/open-telemetry/opentelemetry-cpp-contrib/actions/runs/35622222700/job/106407795774?pr=667) and leaving #668 open until they can be resolved in a future PR?

exporters/prometheus/CMakeLists.txt
thompson-tomo: Fyi this has caused issues in https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/653

thompson-tomo:
@dbarker have raised https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/713 to address the test failing on ubuntu.

I have added a skip on the testing step based on runner os. That should get the ci green.

dbarker (approved):
Thanks for the fixes

PR #669, opened 2026-08-24
bump min cmake in working ci

thompson-tomo:
This bumps the min cmake to 3.16 to support fetchContent.


exporters/fluentd/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.24)
```
dbarker: The cmake minimum should be no higher than 3.22 for now, but ideally we match the minimum 3.16 supported by the otel-cpp core project. 

The rationale is to follow the [Google foundational c++ support policy](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#compilers-tools-build-systems). This policy is followed by many otel-cpp dependencies including: googletest, benchmark, protobuf, grpc, and absl. In `opentelemetry-cpp` the minimum is 3.16 to match the minimum from the latest release of those dependencies.

dbarker (changes requested):
Thanks for the PR. If updating the CMake minimums in this repo, please set all to 3.16. Please see the rationale below.

exporters/fluentd/CMakeLists.txt
```diff
@@ -167,6 +167,7 @@ if(BUILD_TESTING)
       googletest
       GIT_REPOSITORY https://github.com/google/googletest.git
       GIT_TAG 15460959cbbfa20e66ef0b5ab497367e47fc0a04 # v1.12.0
+      FIND_PACKAGE_ARGS 1
```
dbarker: To support older CMake versions we need to follow an explicit find/fetch strategy for dependencies. Once we can upgrade the CMake minimum to 3.24 (which will be in 2027) we can use the built-in feature from FetchContent. 

To avoid replication in all contrib projects the find/fetch strategy for each dependency should be implemented in a separate cmake script (e.g. cmake/googletest.cmake).

See the otel-cpp implementation here: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/googletest.cmake

and the prototype for the contrib repo in: https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/550/changes#diff-66c950bc30e916da2d6233cf678d01cd39cc61992a27f6adf3aaf81f5a568a6f

exporters/fluentd/CMakeLists.txt
thompson-tomo: Done

exporters/fluentd/CMakeLists.txt
thompson-tomo: My understanding is that, the arguments are ignored if they are not supported hence it shouldn't be an issue to have them here.

dbarker: I just tested with 3.16 and adding `FIND_PACKAGE_ARGS 1` to the declare call fails configuration. Upgrading to 3.24 resolves the error. 

```
[1/9] Creating directories for 'googletest-populate'
[1/9] Performing download step (git clone) for 'googletest-populate'
fatal: could not create work tree dir '': No such file or directory
fatal: could not create work tree dir '': No such file or directory
fatal: could not create work tree dir '': No such file or directory
-- Had to git clone more than once:
          3 times.
CMake Error at googletest-subbuild/googletest-populate-prefix/tmp/googletest-populate-gitclone.cmake:31 (message):
  Failed to clone repository: 'https://github.com/google/googletest.git'


FAILED: googletest-populate-prefix/src/googletest-populate-stamp/googletest-populate-download 
cd /home/devuser/build/_deps && /opt/cmake/bin/cmake -P /home/devuser/build/_deps/googletest-subbuild/googletest-populate-prefix/tmp/googletest-populate-gitclone.cmake && /opt/cmake/bin/cmake -E touch /home/devuser/build/_deps/googletest-subbuild/googletest-populate-prefix/src/googletest-populate-stamp/googletest-populate-download
ninja: build stopped: subcommand failed.

CMake Error at /opt/cmake/share/cmake-3.16/Modules/FetchContent.cmake:915 (message):
  Build step for googletest failed: 1
Call Stack (most recent call first):
  /opt/cmake/share/cmake-3.16/Modules/FetchContent.cmake:1006 (__FetchContent_directPopulate)
  /opt/cmake/share/cmake-3.16/Modules/FetchContent.cmake:1047 (FetchContent_Populate)
  cmake/googletest.cmake:34 (FetchContent_MakeAvailable)
  CMakeLists.txt:473 (include)


-- Configuring incomplete, errors occurred!
```

Please revert that FIND_PACKAGE_ARGS change and I'll approve.

thompson-tomo: Thanks for testing, that is strange it was failing. Have reverted the package args.

dbarker (approved):
Thanks for the fix.

Issue #668, opened 2026-08-24
[CI] fix or remove broken workflow jobs


dbarker:
The following GitHub Action workflow jobs are currently broken and were disabled in https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/664.

This issue is to track work to either enable or remove the broken jobs. Please link this issue to PRs that fix or remove these jobs. 

### [webserver.yml](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/.github/workflows/webserver.yml) broken jobs

- [ ] - webserver instrumentation CI / static-analysis (CodeQL)
- [ ] - webserver instrumentation CI / webserver-centos7-build

### [prometheus.yml](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/.github/workflows/prometheus.yml)  broken jobs

- [ ] - prometheus instrumentation CI / CMake on Linux 
- [ ] - prometheus instrumentation CI / Bazel on MacOS 
- [ ] - prometheus instrumentation CI / Bazel on Linux 

### [nginx.yml](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/.github/workflows/nginx.yml)  broken jobs

- [ ] - nginx instrumentation CI / nginx 

### [httpd.yml](https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/.github/workflows/httpd.yml)  broken jobs

- [ ] - httpd instrumentation CI / Build module
- [ ] - httpd instrumentation CI / Build module with CMake
- [ ] - httpd instrumentation CI / Run end-to-end tests

proost:
@dbarker 

Should we continue maintaining centos7 support?

I investigated centos7 issues, centos7 reached EOL(see the https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/461). centos linux has also been discontinued. See the https://www.centos.org/centos-linux-eol/, https://blog.centos.org/2020/12/future-is-centos-stream/

This repository already has an active almalinux8 job providing EL8-compatible build/test coverage and producing the webserver artifact.

I propose:

1. Remove the inactive centos7 CI job.
2. Mark centos7 as deprecated and unsupported in the webserver documentation. 
3. Recommend almalinux8 as the migration target.
4. Retain the existing centos7 docker temporarily, with removal planned for the next major webserver module release.

[...]

PR #679, opened 2026-09-02
refactor: modernizing httpd module

proost:
issue: https://github.com/open-telemetry/opentelemetry-cpp-contrib/issues/668

## Why the CI jobs were failing

1. Bazel jobs — boringssl compiled with `-Werror`
`WORKSPACE` pinned opentelemetry-cpp at commit `b4584ade`. Its transitive gRPC dependency
pulls a boringssl snapshot whose Bazel `copts` include `-Werror`. GCC 11 and GCC 13 added
diagnostics that fire on that boringssl vintage, turning them fatal.

2. CMake job — gRPC v1.36.4's vendored abseil

`setup-cmake.sh` cloned gRPC v1.36.4, whose vendored abseil `lts_2020_09_23` no longer
compiles on either runner.

-----

Following otel-cpp repo, everything is changed.
1. `WORKSPACE` is removed, `MODULE.bazel` and `bazelrc` is introduced.
2. because version is bumped, source code reflects on that.
3. bump otel-cpp 1.28 which is latest version.


instrumentation/httpd/setup-cmake.sh
```diff
@@ -4,72 +4,55 @@ set -euxo pipefail
...
+   -DCMAKE_CXX_STANDARD=14 \
+   -DWITH_OTLP_GRPC=ON \
+   -DBUILD_TESTING=OFF \
+   -DWITH_EXAMPLES=OFF \
+   -DWITH_BENCHMARK=OFF \
+   -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+   ..
+make -j"$(nproc)"
+make install
+cd -
```
thompson-tomo: Can we move this to a fetchContent in the cmake?

instrumentation/httpd/setup-cmake.sh
```diff
@@ -4,72 +4,55 @@ set -euxo pipefail
...
+   -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+   -DgRPC_BUILD_GRPC_NODE_PLUGIN=OFF \
+   -DgRPC_BUILD_GRPC_OBJECTIVE_C_PLUGIN=OFF \
+   -DgRPC_BUILD_GRPC_PHP_PLUGIN=OFF \
+   -DgRPC_BUILD_GRPC_PYTHON_PLUGIN=OFF \
+   -DgRPC_BUILD_GRPC_RUBY_PLUGIN=OFF \
+   ../..
+make -j"$(nproc)"
+make install
+cd -
```
thompson-tomo: Can we move this to a fetchContent in the cmake?

instrumentation/httpd/setup-cmake.sh
proost: [5d8b99b](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/679/commits/5d8b99b5996f7fa847ec363d9b114277b9726325)

.github/workflows/httpd.yml
thompson-tomo: It would be great to replicate the changes in https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/651 that way the codeql workflow can install what is needed

.github/workflows/httpd.yml
```diff
@@ -30,14 +29,18 @@ jobs:
...
+      - name: Setup dependencies
+        working-directory: instrumentation/httpd
         run: |
-          sudo ./instrumentation/httpd/setup-buildtools.sh
-          sudo ./instrumentation/httpd/setup-environment.sh
+          sudo apt update -y
+          xargs -a apt-packages.txt sudo apt install -y --no-install-recommends --no-install-suggests
+      - name: Setup bazel
+        working-directory: instrumentation/httpd
+        run: sudo ./setup-buildtools.sh
```
thompson-tomo: Could we switch to https://github.com/bazel-contrib/setup-bazel

.github/workflows/httpd.yml
```diff
@@ -86,11 +82,15 @@ jobs:
...
         run: |
-          sudo ./instrumentation/httpd/tests/setup-tools.sh
-          sudo ./instrumentation/httpd/setup-buildtools.sh
-          sudo ./instrumentation/httpd/setup-environment.sh
+          sudo apt update -y
+          xargs -a apt-packages.txt sudo apt install -y --no-install-recommends --no-install-suggests
+          xargs -a tests/apt-packages.txt sudo apt install -y --no-install-recommends --no-install-suggests
+      - name: Setup bazel
+        working-directory: instrumentation/httpd
+        run: sudo ./setup-buildtools.sh
```
thompson-tomo: Could we switch to https://github.com/bazel-contrib/setup-bazel

thompson-tomo (approved):

.github/workflows/httpd.yml
```diff
@@ -19,7 +19,6 @@ permissions:
 jobs:
   build:
```
dbarker: Please remove this build job in favor of the super-set e2e test job. There doesn't seem to be a good use case for this build job since the e2e builds and tests the component.

.github/workflows/httpd.yml
```diff
@@ -30,14 +29,21 @@ jobs:
...
+        uses: bazel-contrib/setup-bazel@c5acdfb288317d0b5c0bbd7a396a3dc868bb0f86 # 0.19.0
+        with:
+          module-root: instrumentation/httpd
+          bazelisk-cache: true
+          repository-cache: true
       - name: Compile
         run: |
           cd instrumentation/httpd && ./build.sh
-          cp bazel-out/k8-opt/bin/otel.so /tmp/mod-otel.so
+          cp bazel-bin/otel.so /tmp/mod-otel.so
```
dbarker: We don't need to upload binary artifacts for this component now. Publishing CI or release artifacts needs to be a separate concern and PR.

instrumentation/httpd/CMakeLists.txt
```diff
@@ -1,13 +1,39 @@
...
+
 find_package(Threads REQUIRED)
-find_package(Protobuf REQUIRED)
-find_package(gRPC REQUIRED)
-find_package(CURL REQUIRED)
-find_package(Thrift REQUIRED)
+
+# Use an installed opentelemetry-cpp when one is available, otherwise build it as part of
+# this project.
+find_package(opentelemetry-cpp CONFIG QUIET)
```
dbarker: This pattern to find or fetch otel-cpp will be repeated often. Please move this to a top level cmake directory and include it here. This will align with the pattern from otel-cpp and reduce the review/maintenance overhead as other components adopt it.

dbarker (changes requested):
Thanks! I'm requesting changes below.

.github/workflows/httpd.yml
proost: [d88ab49](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/679/commits/d88ab49418390ea1c86ca5acae835e1ac084fe0c)

instrumentation/httpd/CMakeLists.txt
proost: [5c3a4bf](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/679/commits/5c3a4bf5e6d7470edc28152b4e03f1d4bfa1bd3d)

OK, But how about change httpd only? Later we can change other modules too.

cmake/opentelemetry-cpp.cmake
```diff
@@ -0,0 +1,48 @@
...
+    GIT_REPOSITORY "https://github.com/open-telemetry/opentelemetry-cpp.git"
+    GIT_TAG "${opentelemetry-cpp_GIT_TAG}"
+    GIT_SHALLOW TRUE)
+  set(opentelemetry-cpp_PROVIDER "fetch_repository")
+
+  FetchContent_MakeAvailable(opentelemetry-cpp)
+
+  if(_otel_cpp_prev_build_testing STREQUAL "")
+    unset(BUILD_TESTING CACHE)
+  else()
```
thompson-tomo: I doubt this structure/pattern would be able to be automatically updated using renovate even with a custom manager.

I would prefer that fetch content has the tag explicitly set to the git sha and if you want a different version it should be found by find_package prior to the fetch.

dbarker: Good point. Setting the SHA directly to the GIT_TAG here and removing ${opentelemetry-cpp_GIT_TAG} is a good change to make.

cmake/opentelemetry-cpp.cmake
```diff
@@ -0,0 +1,48 @@
...
+    set(BUILD_TESTING "${_otel_cpp_prev_build_testing}" CACHE BOOL "" FORCE)
+  endif()
+  unset(_otel_cpp_prev_build_testing)
+
+  string(REGEX REPLACE "^v([0-9]+\\.[0-9]+\\.[0-9]+)$" "\\1" opentelemetry-cpp_VERSION
```
dbarker: The opentelemetry-cpp project sets `OPENTELEMETRY_ABI_VERSION_NO` and `OPENTELEMETRY_VERSION` as CMake cache variables that can be used here if the project was found or fetched.

instrumentation/httpd/CMakeLists.txt
```diff
@@ -50,14 +58,14 @@ set_target_properties(otel_httpd_module PROPERTIES
...
   ${APR_INCLUDE_DIR}
 )
 
 target_link_libraries(otel_httpd_module
   PRIVATE
-  ${OPENTELEMETRY_CPP_LIBRARIES}
-  gRPC::grpc++
-#  Thrift::thrift
+  opentelemetry-cpp::api
+  opentelemetry-cpp::trace
```
dbarker, line 68: Thanks for the cleanup to use the CMake targets. This helps clarify what dependencies the component currently needs. It is a bit of a red flag that the instrumented module needs SDK and exporter targets. Are these only needed by tests or examples?

The instrumented libraries in contrib should only depend on the opentelemetry-cpp::api to allow application developers using the instrumentation to decide what SDK components to link.

instrumentation/httpd/CMakeLists.txt
dbarker: Thanks for the update. Agree this PR should only update httpd. Other components can be separate PRs.

cmake/opentelemetry-cpp.cmake
proost: That's good point. [cca68aa](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/679/commits/cca68aa86b2c522cc31b57f8e53adfbef9f7e665)

cmake/opentelemetry-cpp.cmake
proost: [0d1486e](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/679/commits/0d1486e71312c68d89c352c7a040f9e1d38d666d)

instrumentation/httpd/CMakeLists.txt
proost: Not tests or examples.  API-only would need source changes. Do you want to handle in this PR too?

dbarker: Thanks for checking. 

> Do you want to handle in this PR too?

No that will be a bigger refactor and it is unclear how users should configure the SDK with the httpd server. Best left to a future PR.

dbarker (approved):
Looks good. Thanks.

PR #682, opened 2026-09-05
fix: move to ubuntu 22.04

proost:
Current otel-webserver-module covers ubuntu20.04. But this passes EOL. See this: https://ubuntu.com/blog/ubuntu-20-04-lts-end-of-life-standard-support-is-coming-to-an-end-heres-how-to-prepare

### Why not ubuntu 24.04?
ubuntu 24.04 needs huge change. 

First, ubuntu 24.04 has no python2. Current build script gyp with python 2.7. 
Second, current old pinned deps are problem for GCC 13 which is default in ubuntu 24.04. So this makes huge updates.

I want to step by step changes.

 Later, I tried to bump deps. This makes harder to handle OS EOL.

### Explain of changes

* "extern" keyword: Because we use gcc 11 and from gcc10 no-common is default option. 
* bump gRPC:  Because glibc version is changed in ubuntu 22.04, SIGSTKSZ become non-constants. This makes compile error.
* bump google-test: Same with compile error. current version has uninitialized variable exist, and GCC11 retuns error.


marcalff (approved):
LGTM, thanks for the upgrade.

